PR #38, opened 2021-08-03
Fix for invalid password bug on Synapse/Official Matrix Server?

Troplo:
# Description

There is a bug where it says Invalid Password on my Synapse server, so I compared the payload against Element and found that the payload is a little different on Element compared to Cinny, I however am unsure if this causes issues on Dendrite or other Matrix server implementations.

Essentially this could be either a simple bug fix or breaking change.

I've tested this on my self hosted Synapse server, and official/default Matrix server. I'm unsure about Dendrite but I doubt it'll cause issues due to it also working with Element fine.

# Fixes

Updated payload in src/client/action/auth.js to reflect the updated authentication payload that doesn't cause the invalid password message on Synapse servers.

## Type of change

# Checklist:

### DIFF
--- a/src/client/action/auth.js
+++ b/src/client/action/auth.js
@@ -15,7 +15,10 @@ async function login(username, homeserver, password) {
   const client = sdk.createClient({ baseUrl });
 
   const response = await client.login('m.login.password', {
-    user: `@${username}:${homeserver}`,
+    identifier: {
+      type: 'm.id.user',
+      user: username,
+    },
     password,
     initial_device_display_name: cons.DEVICE_DISPLAY_NAME,
   });


ajbura:
How can I reproduce this " Invalid Password"?

Troplo:
If you attempt to login with an account on the official/default Matrix.org server, you should get the error "M_FORBIDDEN: Invalid password" despite you entering the correct login details. It appears to be consistent on matrix.org and my own Synapse server.

ajbura:
Although I am not able to reproduce it, but merging it since [client-server-api](https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-login) deprecate user method in favour of identifier.

PR #45, opened 2021-08-07
Add Typescript support, refactor some components, & more

phildenhoff:
# Description

This monster is probably a larger PR than I would ever want to merge, as it does a number of things. Essentially it was my mainline while I toyed with migrating to the project to Typescript.

It's not a total migration, there are only a few files converted, but I tried to take care to clarify the design (when possible!), removed anti-patterns like [using refs for getting values](https://reactjs.org/docs/refs-and-the-dom.html) or adding styles from JS, and split out parts that are logically separate to make working on the project easier for new devs. My main drive was, essentially, "how do I make this easier (for me and for others) to work on".

General feature list (the commit history should also be a good list):

- Adds Typescript support to the build system
- Migrates `Auth.jsx` to `Auth.tsx` (typescript), changes the order of the file to avoid 'not-yet-defined' warnings
- Adds Formik to manage form state for login / signup
- Moves auth validation logic into an `auth.ts` file
  - I'm planning on expanding upon this to use `yup` (or similar) to remove the hacky validation logic I wrote 😁
- Migrates `Text.jsx` to `Text.tsx`, `Input.jsx` to `Input.tsx`, `IconButton.jsx` to `IconButton.tsx`
- Splits `Input` component into `Input` and `ResizableInput` components
- `Text`, `IconButton`, and `Input` / `ResizableInput` components are no longer exported as default, so they have to be imported as `import { Text }` instead of `import Text` — this is what caused like... most of the files to be changed
  - not a hill I'd die on; happy to change these back to be default exports. Not sure how `Input` would work with two exports though? I guess it becomes two files?
- Adds a helper `Result` type so that we can get one type back from things like `isUserIdValidForLogin` and then interrogate that type to get either a `Ok` result or an `Err`
- The Settings page now shows the device key alongside it's ID — both are important to verify the session from Element, Seaglass, FluffyChat, or whatever
Screen Shot 2021-08-06 at 11 01 12 PM
- Added the `/dist` folder to .gitignore, as well as `yarn.lock` and `.vscode/` because those are the tools I use and I don't want to accidentally commit them 😅
- Changed or disabled some `.eslint` rules that were triggering needlessly
  - In general, I strongly prefer "set and forget" defaults (like what Prettier provides with formatting) but research couldn't help me figure out why rules like `eslint/require-default-props` were triggering for components that _had_ default props, so I opted to disable it. Not a hill I'd care to die on though — I support reinstating any of the rules.

Good lord, that's... yeah even that list is a lot. I'm happy to discuss any of these changes in this PR (or on Matrix if you'd prefer: @phildenhoff:matrix.org), explain my thinking, or make changes. Nothing that I've done, as far as I can tell, has changed the look of the UI at all, but it's definitely worth someone else pulling the branch first to double check. Looking forward to feedback 🙂

## Type of change

# Checklist:


farribeiro:
Needs fix the conflicts, please

kylethorpe:
Needs fix the conflicts, please

phildenhoff:
Haha thank you both for your comments. I'd like to have more of a discussion around the PR before spending more time to resolve merge conflicts. Constantly trying to resolve conflicts while discussing aspects of the PR would be a sisyphean task :)

farribeiro:
If don't v fix the conflicts, so it won't open the merge button

ajbura:
I have given it deep preview and most important thing I am concern about is the addition of typescript. I have no experience with typescript and being a major contributor it going to affect the project, which is the last thing I want. Also I love the flexibility of JavaScript and it might be possible to make typescript as flexible as my need but again as I have no experience with it, and I don't want to spend time learning that.

The main reason I am declining this PR is typescript but if you want to add Formik and other improvements, feel free to open a separate PR for each.

phildenhoff:
No worries, I can appreciate your concerns. Just like you're concerned about Typescript, I don't have much of an interest in using standard Javascript so I won't be adding Formik or the other features I built in this PR. However, feel free to take my work as inspiration if you'd like to add those back in 😄

C0ffeeCode:
I do not think Typescript takes away JavaScript's flexibility but instead makes it even more flexible by allowing us to bring more order to the code and making it more friendly to read and understand while most of it is optional.
Also, (for me) it allows me to write my code faster since my IDE can give me better auto completion etc.
I do not think that a Js developer would really need to learn Ts but since they are so similar and a Js dev should be able to read Typescript as well.

It seems like using Js is preferred by the maintainers but also keeps away interested developers.
Therefore, it might be thinkable to allow TypeScript without the goal to replace existing code.
New contributors could choose to go for Ts and others for JavaScript.

PR #139, opened 2021-10-11
Add bidi support

mskf1383:
### Description

Curentlly Cinny is not supporting RTL texts. With this style Cinny become compatible with both LTR and RTL texts.

I check it using `inspect` on my browser, becaus i can't build Cinny on my system.

#### Type of change

### DIFF
--- a/src/app/atoms/text/Text.scss
+++ b/src/app/atoms/text/Text.scss
@@ -39,3 +39,12 @@
   @include font(b3, 400);
   color: var(--tc-surface-low);
 }
+
+.text-h1, .text-h2,
+.text-s1,
+.text-b1, .text-b2, .text-b3,
+p {
+direction: ltr;
+unicode-bidi: plaintext;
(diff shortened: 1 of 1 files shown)


ajbura:
Cinny already has support for RTL languages, it's just that there's no button to enable it. So adding bidi text support is kind of pointless and also break the RTL.

See the comparison below:

rtl-urdu
Photo: Cinny's RTL support

rtl-urdu-bidi
Photo: Your bidi support

mskf1383:
Beautiful!
But is this support **both** RTL and LTR **at a time**?
Sometimes the text is RTL but has some LTR parts.
Completely RTL:
Screenshot from 2021-10-12 08-52-01
Bidi:
Screenshot from 2021-10-12 08-52-28

The important thing is that style. Please add this to both LTR and RTL style:
```css
unicode-bidi: plaintext;
text-align: initial;
```

Thnaks!

PR #159, opened 2021-11-10
Use Unicode aware character-wise slicing

samueldr:
### Description

`slice` on a String will slice at the byte boundaries. This breaks
avatars for users with emojis or any "complex" unicode characters as the
first character of their names.

This also applies to Rooms.

image → image

#### Type of change

### Checklist:


samueldr:
Considering this is all happening with `<Avatar />`, wouldn't it be better if instead of taking the first char at the call site, the slicing happened *in* the `<Avatar />` component?

ajbura:
@samueldr Yes it will be good if slicing happen in <Avatar /> comp.

ajbura (approved):

PR #160, opened 2021-11-10
Suggest all commands when using `/`

### LINKED ISSUE #157: Show all commands when typing "/"
Kind of annoying that I can't view all the commands available in Cinny. 

Pressing "/" (maybe also a tab) can give a pop up with a list of commands

#### Describe alternatives you've considered
Add available commands in the command button

samueldr:
### Description

This implements a suggestion, that I am in agreement with. Knowing which
commands exists when first using Cinny is cumbersome, as completion
requires at least one letter to be used. I had to poke at arbitrary
letters to see if I was missing any commands at first.

By always suggesting all commands initially it removes the awkwardness
of discovery. At the cost of losing the activation message.

Fixes #157.

**Note**: I'm not sure the code is entirely *correct*. It works, it works as it should. But I'm not sure whether I'm missing some use cases from asyncSearch, or if I'm somewhat misusing event emission.

I'm also unsure whether this is desirable by the project. But eh, this was made mainly because I was playing around in this area and it helped me better understand how this works.

#### Type of change

### Checklist:

### DIFF
--- a/src/app/organisms/room/RoomViewCmdBar.jsx
+++ b/src/app/organisms/room/RoomViewCmdBar.jsx
@@ -163,26 +163,6 @@ FollowingMembers.propTypes = {
   viewEvent: PropTypes.shape({}).isRequired,
 };
 
-function getCmdActivationMessage(prefix) {
-  function genMessage(prime, secondary) {
-    return (
-      <>
-        <span>{prime}</span>
-        <span>{secondary}</span>
-      </>
-    );
-  }
-  const cmd = {
-    '/': () => genMessage('General command mode activated. ', 'Type command name for suggestions.'),
-    '>*': () => genMessage('Go-to command mode activated. ', 'Type space name for suggestions.'),
-    '>#': () => genMessage('Go-to command mode activated. ', 'Type room name for suggestions.'),
-    '>@': () => genMessage('Go-to command mode activated. ', 'Type people name for suggestions.'),
-    ':': () => genMessage('Emoji autofill command mode activated. ', 'Type emoji shortcut for suggestions.'),
-    '@': () => genMessage('Name autofill command mode activated. ', 'Type name for suggestions.'),
-  };
-  return cmd[prefix]?.();
-}
-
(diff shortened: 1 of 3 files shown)


src/app/organisms/room/RoomViewCmdBar.jsx
```diff
@@ -441,30 +441,33 @@ function RoomViewCmdBar({ roomId, roomTimeline, viewEvent }) {
     );
   }
 
+  const inSuggestionMode = cmd !== null;
+  const hasSuggestions = inSuggestionMode && (typeof cmd.suggestions === 'undefined' || cmd.suggestions?.length > 0);
```
samueldr, line 445: This change isn't strictly needed, but it made it much easier to understand the following section.

ajbura:
@samueldr I have not tested it but it is not good to change general functionality of AsyncSearch. Instead this can be easily achieved by returning getCmdSuggestions from getCmdActivationMessage for "/" cmd.

samueldr:
AFAICT returning `getCmdSuggestions` from `getCmdActivationMessage` would require some rework since `fireCmd` and the correct `cmd` reference are defined in the `<RoomViewCmdBar />` component.

Though uh, yeah, if I move `getCmdSuggestions` into the component's scope it seems to want to work...

But the changes to asyncSearch *are* needed. This is because otherwise it's forcibly sending an empty result set on empty string searches:

https://github.com/ajbura/cinny/blob/2c9e32b6c472df133a68d19754ed7e5e9691fc31/src/util/AsyncSearch.js#L60-L63

I updated with (part of) your suggestion.

ajbura:
You can pass `cmd` and `fireCmd` as an argument or can use callback to call `getCmdSuggestions` also `asyncSearch.search` doesn't send empty array cause it never gets called inn `activateCmd`(which gets called when user enter cmd prefix eg.`/`).

ajbura:
This has been implemented along with removing go to commands from message input in recent release.

Issue #83, opened 2021-08-29
Custom Emoji & Stickers using MSC2545


BBaoVanC:
#### Describe the solution you'd like

Currently [MSC2545](https://github.com/matrix-org/matrix-doc/pull/2545) is completely unsupported in Cinny. Emotes just show as the name, and stickers just show as regular images. It would be nice to be able to properly view and send custom stickers and emoji using Cinny.

#### Additional context

MSC2545 is currently supported in Nheko and FluffyChat. It allows you to create sticker/emoji packs which can be added to your user, or to rooms.

ajbura:
Hi, how can I test them in FluffyChat? I have tried to add some custom emojis, but it doesn't seems to wrok?

BBaoVanC:
> Hi, how can I test them in FluffyChat? I have tried to add some custom emojis, but it doesn't seems to wrok?

You can press the plus on the chat box, then "Send sticker", or you can type a colon and then start typing the name of the emote (ex. `:cat` will show custom emoji starting with "cat"). There's also an emote pack in the FluffyChat room with a single FluffyChat sticker, and I also have some in my room ([#general:boba.best](https://matrix.to/#/#general:boba.best)).

jasikpark:
In testing element still doesn't have support, but Nheko Reborn has both room level support, globally adding emoji packs, and uploading your own available

but QT D:

licentiapoetica:
I noticed that the stickers/emojis appear in cinny when you have added them in nheko for example but don't get rendered in the message

PR #161, opened 2021-11-10
Add basic support for displaying emotes

samueldr:
### Description

This only handles *displaying* emote events in a specific manner
recognizable as an emote.

It requires passing the sender name to message content elements, so that
the name can be in-lined into the message.

Sending emotes requires more changes. Namely better support for
commands. It would be required that they're not tied to the
completion framework for activation, but hooked on message being
sent and starting with a `/`. (WIP elsewhere)

**Note**: Not to be confused with "custom emoticons" from #83.

#### Testing

You will need a client that can send emotes (e.g. `/me`) commands.

I was using Element at first, but continued by dogfooding the WIP feature in Cinny.

#### Type of change

### Checklist:

### DIFF
--- a/src/app/molecules/message/Message.jsx
+++ b/src/app/molecules/message/Message.jsx
@@ -106,10 +106,17 @@ MessageReply.propTypes = {
   content: PropTypes.string.isRequired,
 };
 
-function MessageContent({ content, isMarkdown, isEdited }) {
+function MessageContent({
+  senderName,
+  content,
+  isMarkdown,
+  isEdited,
+  msgType,
+}) {
   return (
     <div className="message__content">
       <div className="text text-b1">
+        { msgType === 'm.emote' && `* ${senderName} ` }
         { isMarkdown ? genMarkdown(content) : linkifyContent(content) }
       </div>
       { isEdited && <Text className="message__content-edited" variant="b3">(edited)</Text>}
@@ -121,9 +128,11 @@ MessageContent.defaultProps = {
   isEdited: false,
 };
 MessageContent.propTypes = {
+  senderName: PropTypes.string.isRequired,
   content: PropTypes.node.isRequired,
   isMarkdown: PropTypes.bool,
   isEdited: PropTypes.bool,
+  msgType: PropTypes.string.isRequired,
 };
 
 function MessageEdit({ content, onSave, onCancel }) {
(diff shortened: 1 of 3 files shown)


src/app/molecules/message/Message.jsx
```diff
@@ -228,10 +233,11 @@ MessageOptions.propTypes = {
 
 function Message({
   avatar, header, reply, content, editContent, reactions, options,
+  msgType,
 }) {
   const msgClass = header === null ? ' message--content-only' : ' message--full';
   return (
-    <div className={`message${msgClass}`}>
+    <div className={`message${msgClass} message--type-${msgType.replace('.', '-')}`}>
```
samueldr: Please tell if you have a preferred style for dynamic class names based on data.

Here I'm transforming the event type (`m.text`) to a form usable as a class name part (`m-text`).

src/app/molecules/message/Message.jsx
```diff
@@ -106,10 +106,13 @@ MessageReply.propTypes = {
   content: PropTypes.string.isRequired,
 };
 
-function MessageContent({ content, isMarkdown, isEdited }) {
+function MessageContent({
+  senderName, content, isMarkdown, isEdited, msgType,
+}) {
```
samueldr: Whoops... didn't realize I "re-formatted" this only only partially.

Let's use this as an opportunity to ask: should the full declaration be kept on a single line, or would you prefer one line per (destructured) param? The more params there are, the easier it gets to merge conflicting changes if it's on multiple lines. (That's why I started changing this but oops forgot to finish.)

src/app/molecules/message/Message.jsx
```diff
@@ -228,10 +233,11 @@ MessageOptions.propTypes = {
 
 function Message({
   avatar, header, reply, content, editContent, reactions, options,
+  msgType,
```
samueldr, line 240: Similarly to the question up top: should this be kept on a single line or is this worthy of being split?

src/app/molecules/message/Message.jsx
ajbura: I personally do not prefer dynamic classes, because there can be uncertain class name collisions.

src/app/molecules/message/Message.jsx
ajbura: I use next line when param > 4

src/app/molecules/message/Message.jsx
samueldr: Which pattern would you like to see used here? I can think of a few, some bad, some not great, some good, with different requirements for maintaining.

Note that we probably need to handle `m.notice` too, to completely support the spec, though at the time both `m.text` and `m.notice` will have no additional styles.

ajbura: What I can think of is appending a new class to myClass if msgType is <abc>

ajbura (approved):

PR #207, opened 2021-12-26
[Enhancement] Use jumbo emoji for short emoji-only messages

### LINKED ISSUE #204: [Enhancement] Display emoji-only messages using jumbo emoji
A lot of chat clients include a feature where messages consisting only emoji (and whitespace) are shown using larger versions of the emoji so that you can see them in more detail.  This behavior is present in (at the very least) Discord, Fluffychat, Mattermost, and I'm sure countless others.

A lot of people, myself included, find that this is a handy feature because it makes the emoji immediately more readable by taking advantage of space that wouldn't have been used by text anyway.

Often, this feature is constrained to a certain number of emoji (say, maybe six), and only to messages without any non-whitespace non-emoji characters, to avoid alignment issues.

I'd like to see a feature like this in Cinny.  To propose specifics:
- Messages containing at most six emoji and no non-whitespace characters would be affected
- Both twemoji and custom emoji would be affected
- Affected emoji would be displayed at 64px
- The implementation should not differ significantly from existing implementations in other apps, to preserve intuitivity

The current behavior is simply to display emoji-only messages (and the emoji within) at the same size as they would appear in a standard message.

Cards on the table, I've actually already written an implementation of this here: https://github.com/Alch-Emi/cinny/tree/big-emoji

I foolishly wrote this implementation before reading the Contributing guidelines, which specify that even minor feature changes should go through an issue before a PR is opened, so I'm opening this issue to discuss feature specifics before the PR is opened.  I'd also like to make it clear that I absolutely understand if this feature is rejected or needs significant changes from my implementation, and will be willing to make those changes if need be.

Thank you for your feedback!

alch-emi:
### Description

Closes #204

Emoji are currently displayed using a small, inline size whenever sent in a message.  This pull request adds a special case where a message consists only of up to 6 emoji and whitespace.  For these messages, emoji are displayed at 64px.  This feature is sometimes referred to as "Jumbo Emoji".  For more details about this change and it's motivations, please see issue #204.

image

One thing I wanted to ask before this was merged - in order to make detecting the number of emoji easier, I removed a `<p>` tag around messages without explicit HTML formatting (which can still contain twemoji).  As far as I can tell, this shouldn't break anything, especially because HTML formatted messages are displayed without being wrapped in a `<p>` tag, so I think that this tag was vestigial.  However, I wanted to double check with you, since you have a much better understanding of the codebase than me, and would be more likely to know  if this was relevant.

#### Type of change

### Checklist:

Preview: https://61c9453e4bd53f4dbf189c70--pr-cinny.netlify.app

### DIFF
--- a/src/app/molecules/message/Message.jsx
+++ b/src/app/molecules/message/Message.jsx
@@ -172,14 +172,14 @@ const MessageBody = React.memo(({
   // if body is not string it is a React element.
   if (typeof body !== 'string') return <div className="message__body">{body}</div>;
 
-  const content = isCustomHTML
+  let content = isCustomHTML
     ? twemojify(sanitizeCustomHtml(body), undefined, true, false)
     : twemojify(body, undefined, true);
 
   // Determine if this message should render with large emojis
   // Criteria:
   // - Contains only emoji
-  // - Contains no more than six emoji
+  // - Contains no more than 10 emoji
   let emojiOnly = false;
   if (content.type === 'img') {
     // If this messages contains only a single (inline) image
@@ -191,7 +191,7 @@ const MessageBody = React.memo(({
     const nEmojis = content.filter((e) => e.type === 'img').length;
 
     // Make sure there's no text besides whitespace
-    if (nEmojis <= 6 && content.every((element) => (
+    if (nEmojis <= 10 && content.every((element) => (
       (typeof element === 'object' && element.type === 'img')
       || (typeof element === 'string' && /^\s*$/g.test(element))
     ))) {
@@ -199,9 +199,15 @@ const MessageBody = React.memo(({
     }
   }
 
+  if (!isCustomHTML) {
+    // If this is a plaintext message, wrap it in a <p> element (automatically applying
+    // white-space: pre-wrap) in order to preserve newlines
+    content = (<p>{content}</p>);
+  }
+
   return (
     <div className="message__body">
-      <div className={`text text-b1 ${emojiOnly ? 'emoji-only' : ''}`}>
+      <div className={`text ${emojiOnly ? 'text-h1' : 'text-b1'}`}>
         { msgType === 'm.emote' && (
           <>
             {'* '}
--- a/src/app/molecules/message/Message.jsx
+++ b/src/app/molecules/message/Message.jsx
@@ -174,11 +174,34 @@ const MessageBody = React.memo(({
 
   const content = isCustomHTML
     ? twemojify(sanitizeCustomHtml(body), undefined, true, false)
-    : <p>{twemojify(body, undefined, true)}</p>;
+    : twemojify(body, undefined, true);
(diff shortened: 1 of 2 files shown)


src/app/molecules/message/Message.jsx
```diff
@@ -174,11 +174,34 @@ const MessageBody = React.memo(({
 
   const content = isCustomHTML
     ? twemojify(sanitizeCustomHtml(body), undefined, true, false)
-    : <p>{twemojify(body, undefined, true)}</p>;
+    : twemojify(body, undefined, true);
```
ajbura, line 177: the `<p>` tag was used to apply the `white-space: pre-wrap`  see: (https://github.com/ajbura/cinny/blob/9854f4eb2d4b7fb90f063a29703b8a348de0aafb/src/app/molecules/message/Message.scss#L160)

also it is not used with customHtml is because it wraps the text in appropriate tags automatically.

src/app/molecules/message/Message.jsx
```diff
@@ -174,11 +174,34 @@ const MessageBody = React.memo(({
...
+      || (typeof element === 'string' && /^\s*$/g.test(element))
+    ))) {
+      emojiOnly = true;
+    }
+  }
 
   return (
     <div className="message__body">
-      <div className="text text-b1">
+      <div className={`text text-b1 ${emojiOnly ? 'emoji-only' : ''}`}>
```
ajbura: There are two issues IMO: 
1. I think 64px is too big in size.
2. Also if we use 64px for img then text-b1 line-height should also be 64px otherwise it will slightly displace emoji upward.

IMO we should just replace `text-b1` with `text-h1` which will scale everything automatically inside. but with that emoji height will be around 34px as the `variant h1` text size is 36px.

Also 34px is good height (looks better too) and we can also increase the number of jumbo emoji more than 6.

src/app/molecules/message/Message.scss
```diff
@@ -195,6 +195,11 @@
       opacity: 0;
     }
   }
+
+  & > .emoji-only img {
+    height: 64px !important;
+  }
+
```
ajbura: this can be removed if above suggestion is applied

src/app/molecules/message/Message.jsx
alch-emi: Gotcha!  I'll figure out how to work around that.

src/app/molecules/message/Message.jsx
alch-emi: Sounds good to me!  I'll get that up in just a minute

alch-emi: image

here's how that looks.  I set the emoji limit to be 10 emoji (which also happens to be what fluffychat uses, as i found out).

ajbura (approved):

PR #224, opened 2021-12-31
Add toggle to use browser's preferred color scheme

### LINKED ISSUE #220: Dynamic dark mode switching with the browser
This might be minor, but as there's a bunch of ways to detect from the browser if currently dark mode is "switched on", it'd be nice if there's a way for cinny to also recognise that and switch accordingly.

gpanders:
### Description

This will use dark mode automatically if the browser requests it.

Fixes #220.

Note, I am *not* very experienced in web development, so please feel free to
say this implementation is totally wrong :) But I wanted to at least provide a
patch to get the ball rolling.

#### Type of change

### Checklist:

Preview: https://61d27b7f8b3f270c590ab1df--pr-cinny.netlify.app


src/index.scss
```diff
@@ -290,7 +289,18 @@
...
+.dark-theme {
+  @include dark-mode();
+}
+
+@media (prefers-color-scheme: dark) {
+  .auto-theme {
+    @include dark-mode();
+  }
+}
+
```
gpanders: I modified this to use a `@mixin` with `@include` rather than simply `@extend` because `@extend` apparently does not work inside `@media` queries.

src/index.scss
```diff
@@ -202,8 +202,7 @@
   --bg-surface-extra-low-transparent: hsla(0, 0%, 91%, 0);
 }
 
-.dark-theme,
-.butter-theme {
+@mixin dark-mode() {
   /* background color | --bg-[background type]: value */
```
ajbura, line 206: instead of creating a  mixin we can keep here .dark-theme only.

And @ extend .dark-theme in .butter-theme and .auto-theme.
Benefit of doing this is that scss compiler will not copy all code of .dark-theme to auto and butter theme and instead it will group those classes together.

src/index.scss
```diff
@@ -290,7 +289,18 @@
   --font-secondary: 'Inter', 'Roboto', sans-serif;
 }
 
+.dark-theme {
+  @include dark-mode();
+}
+
+@media (prefers-color-scheme: dark) {
+  .auto-theme {
+    @include dark-mode();
```
ajbura, line 298: this can be replaced with @ extend .dark-theme; if above suggestion is applied.

Not sure how this will work with media query.

src/index.scss
```diff
@@ -290,7 +289,18 @@
...
+  }
+}
+
 .butter-theme {
+  @include dark-mode();
```
ajbura, line 303: this can be replaced with @ extend .dark-theme; if above suggestion is applied.

src/index.scss
```diff
@@ -290,7 +289,18 @@
   --font-secondary: 'Inter', 'Roboto', sans-serif;
 }
 
+.dark-theme {
```
ajbura: .butter-theme can still be grouped with .dark-theme

src/index.scss
ajbura: this can be removed if above suggestion applied

ajbura:
One thing which make it little wired is that `Auto` is not a theme but a behaviour. But current implementation put `Auto` as theme in UI.

Maybe there should be a checkbox or something similar.

gpanders:
> One thing which make it little wired is that `Auto` is not a theme but a behaviour. But current implementation put `Auto` as theme in UI.
> 
> Maybe there should be a checkbox or something similar.

It goes without saying that it's your call of course, but I will just point out that there is precedent for treating "Auto" as a theme unto itself. For example, macOS has "Light", "Dark", and "Auto" themes.

In contrast, GitHub has a separate option for "Sync with system". So I guess both methods are common. Let me know what you want to do and I'll do my best to implement it.

kfiven:
I believe we should go with how GitHub is doing as we also have more than one light/dark themes or do you have anything else in mind @ajbura ?

ajbura:
Yeah I also think we should go with Github choice. cc: @kfiven @gpanders

gpanders:
So what does the UX look like in this case? A checkbox that says “Follow system theme” (or something similar) that, when checked, disables or hides the theme selection buttons and simply uses light or dark mode depending on what the system wants?

gpanders:
Ok I attempted this in 908e189, let me know what you think. If there's a better solution I'm open to feedback, as I said I don't have much experience in this area :)

ajbura (approved):
LGTM! thanks @gpanders :)

PR #252, opened 2022-01-23
Add Desktop notifications

### LINKED ISSUE #182: Desktop notifications
When I receive a notification, I am not actually notified unless I am looking at the Cinny app tab in my browser.

Many web apps leverage browser functionality to send notifications to the desktop, allowing native popup notifications when an event occurs. This requires additional browser permissions for the tab.

#### Describe alternatives you've considered
A native desktop app would require more development effort

#### Additional context
N/A

ginnyTheCat:
### Description

This adds Desktop notification support.

image
image

(These screenshots are taken in GNOME. They will look different on Windows and macOS.)

This change will however not play a custom sound like element web does. The OS will play a generic notification sound on it's own. Adding this feature at a later date however wouldn't be difficult.

Currently Cinny does not ask for notification permission by default on first launch like element web does. The permission can be granted by going to `settings > notifications > request permission`. Personally I think the element approach is more intuitive for new users. If the maintainers are ok with Cinny asking for notification permission after the first login I am more than happy to add this.

Fixes #182

#### Type of change

### Checklist:

Preview: https://61f417e0a83c150ef05c7d85--pr-cinny.netlify.app


ajbura:
> Personally I think the element approach is more intuitive for new users. If the maintainers are ok with Cinny asking for notification permission after the first login I am more than happy to add this.

I am ok with asking notification permission after login.

Also I am getting the following error with crash when opening notifications tab in settings in Safari:
`undefined is not an object (evaluating 'window.Notification.permission')`

ginnyTheCat:
It should not crash in Safari now and asks for permission on startup.

src/client/state/Notifications.js
```diff
@@ -158,6 +167,34 @@ class Notifications extends EventEmitter {
...
+      title = room.name;
+    } else if (mEvent.sender) {
+      title = `${mEvent.sender.name} (${room.name})`;
+    }
+
+    const noti = new window.Notification(title, {
+      body: mEvent.getContent().body,
+      icon: mEvent.sender?.getAvatarUrl(this.matrixClient.baseUrl, 36, 36, 'crop'),
+    });
+    noti.onclick = () => selectRoom(room.roomId);
```
ajbura: We may pass the eventId to selectRoom as second argument to focus that event after room gets open.

src/app/hooks/usePermission.js
```diff
@@ -0,0 +1,24 @@
...
+  useEffect(() => {
+    navigator.permissions.query({ name }).then((_descriptor) => {
+      descriptor = _descriptor;
+
+      update();
+      descriptor.addEventListener('change', update);
+    });
+
+    return () => descriptor.removeEventListener('change', update);
+  }, []);
```
ajbura: I think this is causing error with crash in safari `undefined is not an object (evaluating 'navigator.permissions.query')`.

also `descriptor` should be defined inside useEffect and while removing eventlistener we should make sure that we are not calling it on `undefined`.

src/client/state/userActivity.js
```diff
@@ -0,0 +1,31 @@
...
+    window.addEventListener('mousedown', activityEvent);
+    // window.addEventListener('mousemove', activityEvent);
+    window.addEventListener('keydown', activityEvent);
+    window.addEventListener('wheel', activityEvent, { passive: true, capture: true });
+    window.addEventListener('focus', activityEvent);
+    window.addEventListener('blur', inactivityEvent);
+  }
+
+  recentlyActive() {
+    return Date.now() - this.lastActive <= 2 * 60 * 1000;
```
ajbura: Not sure what 2 * 60 * 1000 means. maybe store it in a const

src/client/state/userActivity.js
```diff
@@ -0,0 +1,31 @@
...
+
+    this._listenEvents();
+  }
+
+  _listenEvents() {
+    const activityEvent = () => { this.lastActive = Date.now(); };
+    const inactivityEvent = () => { this.lastActive = 0; };
+
+    window.addEventListener('mousedown', activityEvent);
+    // window.addEventListener('mousemove', activityEvent);
```
ajbura: please remove if code is not in use

src/client/state/Notifications.js
```diff
@@ -172,6 +209,10 @@ class Notifications extends EventEmitter {
 
       const noti = this.getNoti(room.roomId);
       this._setNoti(room.roomId, total - noti.total, highlight - noti.highlight);
+
+      if (this._isSyncing) {
```
ajbura: `this.matrixClient.getSyncState()` can be used to know syncing state.

src/client/state/Notifications.js
```diff
@@ -190,6 +231,14 @@ class Notifications extends EventEmitter {
...
       }
     });
+
+    this.matrixClient.on('sync', (state) => {
+      if (state === 'SYNCING') {
+        this._isSyncing = true;
+      } else if (state === 'STOPPED' || state === 'ERROR') {
+        this._isSyncing = false;
+      }
+    });
```
ajbura: this code can be removed if `this.matrixClient.getSyncState()` is used at line: 213. I am not sure what else purpose it serve.

src/client/state/Notifications.js
```diff
@@ -158,6 +167,34 @@ class Notifications extends EventEmitter {
     [...parentIds].forEach((parentId) => this._deleteNoti(parentId, total, highlight, roomId));
   }
 
+  async _displayPopupNoti(mEvent, room) {
+    if (!settings.showNotifications) return;
+    if (window.Notification === undefined || window.Notification.permission !== 'granted') return;
```
ajbura: if someone deny the `window.Notification.permission` why `settings.showNotifications` flag it as true? may be we should also update  `settings.showNotifications` as soon as user reset the permission so we can have a one source of truth.

src/client/state/settings.js
```diff
@@ -110,6 +111,15 @@ class Settings extends EventEmitter {
     return settings.isPeopleDrawer;
   }
 
+  getShowNotifications() {
+    if (typeof this.showNotifications === 'boolean') return this.showNotifications;
+
+    const settings = getSettings();
+    if (settings === null) return true;
+    if (typeof settings.showNotifications === 'undefined') return true;
```
ajbura, line 124: if user has not allow notification permission then setting showNotification to true is a bad idea

src/client/state/Notifications.js
ginnyTheCat: I thought about a users disabling notifications, removing the permission and adding it again. After that notifications should still disabled in the settings. Maybe however it would be better to enable them again. I'm not sure.

ajbura: it's better to auto enable them again but if they don't grant permission again `settings. showNotifications` still have a true value which is wrong. And can lead to bugs

ginnyTheCat: `settings.showNotifications` is more like a do-not-disturb setting. The only bug that could occur would be sending notifications when `window.Notification.permission` is not `"granted"` which would error out at `new Notification`.
When updating `settings.showNotifications` based on browser-side permissions the value would still be wrong until a new notification arrives (we would have to do the check either way -> would not catch the bug above in most circumstances).

src/client/state/userActivity.js
```diff
@@ -1,5 +1,7 @@
 import EventEmitter from 'events';
 
+const recentlyActiveThreshold = 2 * 60 * 1000; // 2 minutes
```
ajbura: pls keep const all caps RECENTLY_ACTIVE_THRESHOLD

src/client/state/Notifications.js
ajbura: what exactly `settings.showNotifications` is used for?

src/app/hooks/usePermission.js
ajbura: This is still crashing on Safari. :D

src/client/state/Notifications.js
ginnyTheCat: It's the notification slider in the settings:
image

When the user has not given Cinny notification permissions the button shown below. This however does not change the `settings.showNotifications`.
image

src/client/state/Notifications.js
```diff
@@ -158,6 +165,34 @@ class Notifications extends EventEmitter {
...
+
+    const actions = this.matrixClient.getPushActionsForEvent(mEvent);
+    if (!actions?.notify) return;
+
+    if (navigation.selectedRoomId === room.roomId
+      && userActivity.recentlyActive()) return;
```
ajbura: `document.visibilityState === 'hidden'` will be more viable instead of `userActivity.recentlyActive()`.

src/client/state/Notifications.js
ajbura: As per my understanding the `settings.showNotifications` flag helps to turn ON/OFF desktop notification.
1. When it is OFF it will flag us to not show desktop notifications (even when browser has `window.Notification.permission === granted` ).
2. But when it's ON it flag us to show desktop notification (even when browser has `window.Notification.permission === denied` ). And this is coded as default case + when Notificaiton.permission is removed by browser. It's the another thing that if above case come user will see blue Request permission button instead of toggle but under the hood it's still on.
3. The above case can lead to bug if third person build something on `settings.showNotifications` considering it as a pure flag to user choice of desktop notification.

We may consider it false by default (if `Notification.permission === denied`) else otherwise and we must set it to false if `Notification.permission` gets denied :)

src/app/organisms/settings/Settings.jsx
```diff
@@ -111,6 +118,46 @@ function AppearanceSection() {
...
+      >
+        Request permission
+      </Button>
+    );
+  };
+
+  return (
+    <div className="set-notifications settings-content">
+      <SettingTile
+        title="Show notifications"
```
ajbura: Also shouldn’t ‘Show desktop notifications’ be more correct name for button?

src/client/state/Notifications.js
ginnyTheCat: `document.visibilityState` is not as reliable but should less code than `userActivity.recentlyActive`

src/app/organisms/settings/Settings.jsx
ginnyTheCat: Element web calls them "desktop notifications". Element android calls them "notifications". Maybe calling them "System notifications" or "Native notifications" would make sense?

src/client/state/Notifications.js
ginnyTheCat: There is no real way to check for the permission getting removed. On Chrome and Firefox we can do that. On Safari however there is no live permission access. That would result in the bug described in 3. still occuring until a notification arrives. Would renaming the settings and adding a comment explaining it make sense?

ajbura: For safari the whole desktop notification system doesn't work so we can keep that option to false by default.

For the browser that support it 
1. by default in `src/client/state/settings.js line 118, 119` we can set it as `window.Notification.permission === 'granted` (if Notification.permission is not support it will be false) 
2. we can put a listener in `src/client/event/settings.js` which will update this if browser permission changes. (src/client/event is a place to listen event and do actions accordingly)

:)

src/client/state/Notifications.js
ajbura: yeah `document.visibilityState` will code less and does almost same as UserActivity class?

src/client/state/Notifications.js
ginnyTheCat: Safari should actually have [notifications support](https://caniuse.com/notifications). It only lacks [permission api support](https://caniuse.com/permissions-api) which would be needed if we want to listen to permissions changing instead of checking them when sending a notification.

src/app/organisms/settings/Settings.jsx
ajbura: I think Desktop notifications is much popular term and the one we should go with.

src/client/state/Notifications.js
ajbura: other solution: we can make settings._showNotification private and instead use a get function as public api which will consider window.Notification.permission while returning (&& of both)

ginnyTheCat: Sounds like a good solution.

ajbura (approved):

PR #263, opened 2022-01-26
Normalize unicode search

ginnyTheCat:
### Description

Add option to `AsyncSearch.js` to normalize unicode strings.

image

#### Type of change

### Checklist:

Preview: https://61f2112cb6db812b3b713a34--pr-cinny.netlify.app

### DIFF
--- a/src/util/AsyncSearch.js
+++ b/src/util/AsyncSearch.js
@@ -15,6 +15,7 @@ class AsyncSearch extends EventEmitter {
     this.searchKeys = null;
     this.isContain = false;
     this.isCaseSensitive = false;
+    this.normalizeUnicode = true;
     this.ignoreWhitespace = true;
     this.limit = null;
     this.findingList = [];
@@ -55,8 +56,7 @@ class AsyncSearch extends EventEmitter {
   search(term) {
     this._softReset();
 
-    this.term = (this.isCaseSensitive) ? term : term.toLocaleLowerCase();
-    if (this.ignoreWhitespace) this.term = this.term.replaceAll(' ', '');
+    this.term = this._normalize(term);
     if (this.term === '') {
       this._sendFindings();
       return;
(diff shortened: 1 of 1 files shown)


src/util/AsyncSearch.js
```diff
@@ -113,13 +115,18 @@ class AsyncSearch extends EventEmitter {
...
-    let myItem = (this.isCaseSensitive) ? item : item.toLocaleLowerCase();
-    if (this.ignoreWhitespace) myItem = myItem.replaceAll(' ', '');
-
+    const myItem = this._normalize(item);
     if (this.isContain) return myItem.indexOf(this.term) !== -1;
     return myItem.startsWith(this.term);
   }
 
+  _normalize(item) {
+    let myItem = item.normalize(this.normalizeUnicode ? 'NFKC' : undefined);
```
ajbura: What does this line do?

ginnyTheCat: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize
If `this.normalizeUnicode` is true it turns characters like `ﬀ` into `ff` or `Ⓓ` into `D` which makes it easier to search.
In the screenshot above it turns `𝓙𝓸𝓷𝓪𝓼` into `Jonas`. It however does not turn things like chinese characters into some kind of ASCII representation.
If `this.normalizeUnicode` is false it makes sure all the same characters use the same unicode codepoint. Otherwise `'\u00F1'` and `'\u006E\u0303'` which are both the character `ñ` would not be equal.

ajbura: Sound good. btw if `this.normalizeUnicode` is false it's looks an unnecessary call `item.normalize(undefined)`

ginnyTheCat:
```js
let string1 = '\u00F1';            // ñ
let string2 = '\u006E\u0303';      // ñ
// they both are the same character

console.log(string1 === string2); // false
// they are only encoded differently
// `string1` is a "n with tilde" (composed form; 1 byte)
// `string2` is a "n" with "tilde" (decomposed form; 2 bytes)

let string1normalized = string1.normalize(undefined);
let string2normalized = string2.normalize(undefined);
// now they're both in their composed form

console.log(string1normalized === string2normalized); // true
```

Maybe the homeserver already normalizes this. Won't hurt doing it anyway.
Replacing `undefined` with `'NFC'` (the default value if param is undefined) might make it more obvious the function is still doing something.

ajbura (approved):

PR #373, opened 2022-03-10
Add recently used section to emoji board

### LINKED ISSUE #194: Usability: Store recently used emojis
Most users use at most 4-5 emojis frequently (e.g. thumbs up, thumbs down, party pinata, rocket, etc.)
These should be easy to reach in the emoji picker.
- Store recently used emojis or
- Have a quick access section, like Element

ginnyTheCat:
### Description

Adds a recently used section (like element web has) to the emoji board.
Currently the icon in the sidebar is a heart. Personally I would think a clock would make for a better fit. I'm not a good designer though so I would leave the icon choice to @ajbura.

image

Fixes #194

#### Type of change

### Checklist:

Preview: https://62331f8e1ac2861711b137df--pr-cinny.netlify.app

### DIFF
--- a/src/app/organisms/emoji-board/EmojiBoard.jsx
+++ b/src/app/organisms/emoji-board/EmojiBoard.jsx
@@ -21,6 +21,7 @@ import Input from '../../atoms/input/Input';
 import ScrollView from '../../atoms/scroll/ScrollView';
 
 import SearchIC from '../../../../public/res/ic/outlined/search.svg';
+import HeartIC from '../../../../public/res/ic/outlined/heart.svg';
 import EmojiIC from '../../../../public/res/ic/outlined/emoji.svg';
 import DogIC from '../../../../public/res/ic/outlined/dog.svg';
 import CupIC from '../../../../public/res/ic/outlined/cup.svg';
@@ -193,6 +194,8 @@ function EmojiBoard({ onSelect, searchRef }) {
   const [availableEmojis, setAvailableEmojis] = useState([]);
   const [recentEmojis, setRecentEmojis] = useState([]);
 
+  const recentOffset = recentEmojis.length > 0 ? 1 : 0;
+
   useEffect(() => {
     const updateAvailableEmoji = (selectedRoomId) => {
       if (!selectedRoomId) {
@@ -237,7 +240,7 @@ function EmojiBoard({ onSelect, searchRef }) {
     const $emojiContent = scrollEmojisRef.current.firstElementChild;
     const groupCount = $emojiContent.childElementCount;
     if (groupCount > emojiGroups.length) {
-      tabIndex += groupCount - emojiGroups.length - availableEmojis.length;
+      tabIndex += groupCount - emojiGroups.length - availableEmojis.length - recentOffset;
     }
     $emojiContent.children[tabIndex].scrollIntoView();
   }
@@ -279,13 +282,21 @@ function EmojiBoard({ onSelect, searchRef }) {
       </div>
       <ScrollView invisible>
         <div className="emoji-board__nav">
+          {recentEmojis.length > 0 && (
+            <IconButton
+              onClick={() => openGroup(0)}
+              src={HeartIC}
+              tooltip="Recent"
+              tooltipPlacement="right"
+            />
+          )}
           <div className="emoji-board__nav-custom">
             {
               availableEmojis.map((pack) => {
                 const src = initMatrix.matrixClient.mxcUrlToHttp(pack.avatar ?? pack.images[0].mxc);
                 return (
                   <IconButton
-                    onClick={() => openGroup(pack.packIndex)}
+                    onClick={() => openGroup(recentOffset + pack.packIndex)}
                     src={src}
                     key={pack.packIndex}
                     tooltip={pack.displayName}
@@ -309,7 +320,7 @@ function EmojiBoard({ onSelect, searchRef }) {
                 [7, FlagIC, 'Flags'],
               ].map(([indx, ico, name]) => (
                 <IconButton
-                  onClick={() => openGroup(availableEmojis.length + indx)}
+                  onClick={() => openGroup(recentOffset + availableEmojis.length + indx)}
                   key={indx}
                   src={ico}
                   tooltip={name}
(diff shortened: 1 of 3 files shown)


src/app/organisms/emoji-board/recent.js
```diff
@@ -0,0 +1,31 @@
+import initMatrix from '../../../client/initMatrix';
+import { emojis } from './emoji';
+
+const eventType = 'io.element.recent_emoji';
+
+function get() {
```
ajbura: this function name need be more descriptive

src/app/organisms/emoji-board/recent.js
```diff
@@ -0,0 +1,31 @@
...
+  const recent = get();
+  const i = recent.findIndex(([u]) => u === unicode);
+  let entry;
+  if (i < 0) {
+    entry = [unicode, 1];
+  } else {
+    [entry] = recent.splice(i, 1);
+    entry[1] += 1;
+  }
+  recent.unshift(entry);
```
ajbura, line 32: using `recent.push` if we are adding new emoji and directly updating the `recent[i][1] += 1;` will be faster approach.

src/app/organisms/emoji-board/recent.js
```diff
@@ -0,0 +1,31 @@
...
+
+function get() {
+  return initMatrix.matrixClient.getAccountData(eventType).getContent().recent_emoji ?? [];
+}
+
+export function getRecentEmojis(limit) {
+  return get()
+    .sort((a, b) => b[1] - a[1])
+    .slice(0, limit)
+    .map(([unicode]) => emojis.find((e) => e.unicode === unicode));
```
ajbura: you are using `map` here, not sure if you are intended to filter it? cuz custom emojis crash the app when EmojiBoard opens

src/app/organisms/emoji-board/EmojiBoard.jsx
```diff
@@ -147,8 +150,9 @@ function EmojiBoard({ onSelect, searchRef }) {
   function selectEmoji(e) {
     if (isTargetNotEmoji(e.target)) return;
 
-    const emoji = e.target;
-    onSelect(getEmojiDataFromTarget(emoji));
+    const emoji = getEmojiDataFromTarget(e.target);
+    onSelect(emoji);
+    addRecentEmoji(emoji.unicode);
```
ajbura: we should not add custom emoji to recent for now as they are not used by wide variety of client. also we are updating `io.element.recent_emoji` which doesn't support custom emoji.

src/app/organisms/room/RoomViewCmdBar.jsx
```diff
@@ -237,6 +238,7 @@ function RoomViewCmdBar({ roomId, roomTimeline, viewEvent }) {
       viewEvent.emit('cmd_fired');
     }
     if (myCmd.prefix === ':') {
+      if (!myCmd.result.mxc) addRecentEmoji(myCmd.result.unicode);
```
ajbura, line 241: same issue as above. we should not add custom emoji to recent for now as they are not used by wide variety of client. also we are updating `io.element.recent_emoji` which doesn't support custom emoji.

ajbura (changes requested):

ajbura:
Ok I'll make a 🕓 icon

src/app/organisms/emoji-board/recent.js
ginnyTheCat: Element web expects the emojis to be sorted by last used. That's we we do this dance of removing it from the array and adding it back on

ajbura: that's ok

src/app/organisms/emoji-board/recent.js
ginnyTheCat: Now that fixed the custom emojis being added, should I still filter it?

ajbura: is it mapped to get emojidata? just wanna to make sure what if `emojis.find` returns `undefined`

ginnyTheCat: Yes, that's the idea. Maybe there is a better way to do it. Should I change `emojis` to be some kind of map to make it more efficient?

ajbura: we can just filter it for now. we will improve things later when working specifically on emojiboard

ginnyTheCat: Not sure if there's still a need to filter now as custom emojis are not being added anymore. Should I still filter just to be sure?

ajbura: yeah, as cinny is not the only client which can update that :)

src/app/organisms/emoji-board/recent.js
```diff
@@ -8,10 +8,15 @@ function getRecentEmojisRaw() {
...
+  const res = [];
+  getRecentEmojisRaw()
     .sort((a, b) => b[1] - a[1])
-    .slice(0, limit)
-    .map(([unicode]) => emojis.find((e) => e.unicode === unicode));
+    .find(([unicode]) => {
+      const emoji = emojis.find((e) => e.unicode === unicode);
+      if (emoji) return res.push(emoji) >= limit;
+      return false;
+    });
```
ajbura, line 18: I am not sure what's happening here. we are using `find` but it's doing `forEach` work. also line 16 is confusing to me can you please explain it?

ghost:
Adding a separator `(<hr />)` between hearth and smiley icon would be better. Otherwise someone may think hearth icon is a different emoji section or something.

ajbura:
Heart icon going to be replaced by recent-clock icon

src/app/organisms/emoji-board/recent.js
ginnyTheCat: The `find` works as a `forEach` but breaks when `true` is being returned. That happens when after an emoji gets inserted into the list it hits the limit (of 3 * 7 emojis).

`res.push()` returns the new array length. It would be equivalent to writing:
```js
if (emoji) {
  res.push(emoji);
  return res.length >= limit;
}
```

ajbura (approved):

PR #377, opened 2022-03-14
Fix view source shows original event for an edited message

### LINKED ISSUE #376: View Source does not show the latest event when message is edited
#### Description
If I send a message and then modify it, the protocol sends two events with the `edit event` having a `m.relates_to` and `m.new_content`. It is upto the UI to render the message as edited (Which Cinny does). 
But if I click on `View Source` it is expected to view the latest event. In Cinny we are shown the original event and not the latest event when we view `View Source`.

#### Example
This is what `Cinny`` shows
```json
{
    "content": {
      "body": "I'm about to edit this message.",
      "msgtype": "m.text",
      "org.matrix.msc1767.text": "I'm about to edit this message."
    },
    "type": "m.room.message",
    "event_id": "$epV0V7xZ0cEfVy7tGgmTwGpKZCC3sGl9D4ey4AxxCWY",
  }
```

This is what `element`` shows
```json
{
  "content": {
    "body": " * I've edited the message.",
    "m.new_content": {
      "body": "I've edited the message.",
      "msgtype": "m.text",
      "org.matrix.msc1767.text": "I've edited the message."
    },
    "m.relates_to": {
      "event_id": "$epV0V7xZ0cEfVy7tGgmTwGpKZCC3sGl9D4ey4AxxCWY",
      "rel_type": "m.replace"
    },
    "msgtype": "m.text",
    "org.matrix.msc1767.text": " * I've edited the message."
  },
  "type": "m.room.message",
  "event_id": "$0oItExKSP_KUYIVqqGTFTWAh5vNB1WdD3lwUlochmYs",
}
```

#### To Reproduce
1. Send a message
2. Edit it
3. Click `View Source` for the edited message

#### Expected behavior
Show the latest event in `View Source`

#### Version
Cinny v1.8.0

clmnin:
Signed-off-by: Clament John <cj@hackerlab.in>

fixes #376 

### Description

When we click view source for an edited message we were showing the original event (the unedited event) instead of the latest edited event.

#### Type of change
- Bug fix (non-breaking change which fixes an issue)

### Checklist:

Preview: https://6230529cace91206d67467a0--pr-cinny.netlify.app

### DIFF
--- a/src/app/molecules/message/Message.jsx
+++ b/src/app/molecules/message/Message.jsx
@@ -469,6 +469,18 @@ function isMedia(mE) {
   );
 }
 
+// if editedTimeline has mEventId then pass editedMEvent else pass mEvent to openViewSource
+function handleOpenViewSource(mEvent, roomTimeline) {
+  const eventId = mEvent.getId();
+  const { editedTimeline } = roomTimeline ?? {};
+  let editedMEvent;
+  if (editedTimeline?.has(eventId)) {
+    const editedList = editedTimeline.get(eventId);
+    editedMEvent = editedList[editedList.length - 1];
+  }
+  openViewSource(editedMEvent !== undefined ? editedMEvent : mEvent);
+}
+
 const MessageOptions = React.memo(({
   roomTimeline, mEvent, edit, reply,
 }) => {
@@ -516,7 +528,7 @@ const MessageOptions = React.memo(({
             </MenuItem>
             <MenuItem
               iconSrc={CmdIC}
-              onClick={() => openViewSource(mEvent)}
+              onClick={() => handleOpenViewSource(mEvent, roomTimeline)}
             >
               View source
             </MenuItem>


src/app/molecules/message/Message.jsx
```diff
@@ -734,7 +735,7 @@ function Message({
         {roomTimeline && !isEditing && (
           <MessageOptions
             roomTimeline={roomTimeline}
-            mEvent={mEvent}
+            mEvent={editedMEvent !== undefined ? editedMEvent : mEvent}
```
ajbura: changing this mEvent is not a good idea as other option in menu also depend on this. If we want to get the edited message we can get it from `roomTimeline.editedTimeline` inside MessageOptions

clmnin: I've moved the changes to `MessageOptions`. Please review the changes.

src/app/molecules/message/Message.jsx
```diff
@@ -479,6 +479,16 @@ const MessageOptions = React.memo(({
   const myPowerlevel = room.getMember(mx.getUserId())?.powerLevel;
   const canIRedact = room.currentState.hasSufficientPowerLevelFor('redact', myPowerlevel);
   const canSendReaction = room.currentState.maySendEvent('m.reaction', mx.getUserId());
+  // check if this has an edited event in the timeline, if so, then replace
+  // View Source with editedMEvent
+  const eventId = mEvent.getId();
+  const { editedTimeline } = roomTimeline ?? {};
+  const isEdited = roomTimeline ? editedTimeline.has(eventId) : false;
```
ajbura: this line can give an error if editedTimeline is undefined. I think we can just use `if (editedTimeline?.has(eventId))`

src/app/molecules/message/Message.jsx
```diff
@@ -479,6 +479,16 @@ const MessageOptions = React.memo(({
   const myPowerlevel = room.getMember(mx.getUserId())?.powerLevel;
   const canIRedact = room.currentState.hasSufficientPowerLevelFor('redact', myPowerlevel);
   const canSendReaction = room.currentState.maySendEvent('m.reaction', mx.getUserId());
+  // check if this has an edited event in the timeline, if so, then replace
+  // View Source with editedMEvent
+  const eventId = mEvent.getId();
```
ajbura: Please use `mEvent.getId()` directly as event id can change

src/app/molecules/message/Message.jsx
```diff
@@ -516,7 +526,7 @@ const MessageOptions = React.memo(({
             </MenuItem>
             <MenuItem
               iconSrc={CmdIC}
-              onClick={() => openViewSource(mEvent)}
+              onClick={() => openViewSource(editedMEvent !== undefined ? editedMEvent : mEvent)}
```
ajbura: creating `handleOpenViewSource` function and doing all editedMEvent finding stuff in that func and opening view source from there will be a good idea cuz then we are doing calc when we actually need them

ajbura:
Also please use comments when absolutely necessary otherwise they create additional burden in maintenance.

src/app/molecules/message/Message.jsx
clmnin: Does react do this calc `editedMEvent !== undefined ? editedMEvent : mEvent` for each rerender? I didn't quite get why you are suggesting us to create a new function `handleOpenViewSource`.

Would be cleaner. But my question is on if we are doing unnecessary calc

ajbura: I mean execution from line 484 to 490 on only require when user click menu item.

so we can do that in `handleOpenViewSource` and call `openViewSource` from there

ajbura (approved):

Issue #55, opened 2021-08-15
Make the app responsive


ajbura:
#### Is your feature request related to a problem? Please describe.
No

#### Describe the solution you'd like
Responsive interface for different devices.

farribeiro:
I think this issue resolves #17

daemonspring:
This is also relevant to desktop users. I have a second monitor in a portrait orientation (1080x1920) and it'd be the ideal place to move the chat to. Cinny doesn't handle that limited width at the moment.

samueldr:
Collapsing the left panel would help, I think. I'm slowly working out *opinionated* changes in the user interface for better information density, and have been playing with a scheme where the left sidebar is collapsed until hovered over.

![image](https://user-images.githubusercontent.com/132835/140995219-30e3ba2b-0f70-4bad-81cf-b43009beb009.png)

(This screenshot includes more WIP features than strictly the collapsible side-bar.)

I *think* it shouldn't be too hard to cleanup and add as a setting option. Though at the current time it's not ready for general use.

If you want to try it out, you can use the following:

 - https://gist.github.com/8bf14929d5b550b0e5ea86ed31379d33

It *may* work as a stylus style if you use the options to pre-process with sass.

N-R-K:
> have been playing with a scheme where the left sidebar is collapsed until hovered over.

I personally don't like this type of UI. It's one of the things I *hated* about element. When I collapse the sidebar, I want it to *stay* collapsed. But element opens it up on window resize and I have to manually collapse it again.

samueldr:
> I personally don't like this type of UI. It's one of the things I _hated_ about element. When I collapse the sidebar, I want it to _stay_ collapsed. But element opens it up on window resize and I have to manually collapse it again.

Note that it stays collapsed even when the window is expanded. It's only *when interacting with the sidebar* that it expands. Note that in this WIP implementation it expands *without disrupting the layout*, there is no reflow of the window contents.

I think I have the same grievances against element's "collapsible" panels. I was always fighting against them. Always wanting them to stay collapsed when they didn't. And also importantly, forced to uncollapse for complex interactions.

Note that the expected implementation would make this *optional*, so people that don't like collapsed panels shouldn't be affected by default.

vwkd:
Any updates on this? Currently Cinny is unusable on mobile.

[...]

PR #401, opened 2022-03-18
Adapt to different device widths

C0ffeeCode:
### Description
Cinny should be more usable on smaller devices or smaller windows.
Therefore, it would be useful in those cases to hide the people drawer since it only takes away too much screen and at some point, hide either the navigation or room view.
The people list can still be accessed on the room settings.
If the UI is separated by nav and room view, a button appears on a room header to go back.

UI on 946px width:
image
UI on 890px width:
image
UI on 400px width on a room:
image
UI on 400px width on navigation:
image

Relates to #55

### What does this not address? (yet)
This PR could be merged before the following points are finished.

- Tab in room settings ([will not be changed](https://matrix.to/#/!rgQbWTfsoUcbvEDAvs:matrix.org/$8kLII94rTcIG7m_vjuD5sDTtJhzKmbo5VyjdiAq1MAQ?via=matrix.org&via=envs.net&via=tchncs.de))
  - replacing scroll view might be considerable
- [x] Settings do not adapt
  - Changes made, [also see](https://github.com/ajbura/cinny/pull/401#issuecomment-1073252886)
- Important UI sometimes gets cut away (to be fixed in another PR)
  - image
- Context menus do not fit (to be fixed in another PR)
  - image

#### Type of change

### Checklist:

  - [x] fix bug: Emoji picker does not appear when width is smaller than 360px

Preview: https://625fcf0d375e065d3faff09a--pr-cinny.netlify.app


iambumblehead:
@C0ffeeCode the screenshots look awesome. I look forward to using your changes :)

ajbura:
Hi @C0ffeeCode, I am doing some design changes in user setting panel, so I'd say wait on that so you don't have to do extra work.

C0ffeeCode:
Ok. 👍 
Does this also include changes to space settings and the TitleWrapper on the Header?
image

C0ffeeCode:
Ok. 👍 
Does this also include changes to space settings and the TitleWrapper ?
image

ajbura:
No, only in user settings.

src/app/organisms/room/Room.jsx
```diff
@@ -48,7 +48,9 @@ function Room() {
         <RoomSettings roomId={roomTimeline.roomId} />
         <RoomView roomTimeline={roomTimeline} eventId={eventId} />
       </div>
-      { isDrawer && <PeopleDrawer roomId={roomTimeline.roomId} />}
+      <div className="room__drawer">
+        {isDrawer && <PeopleDrawer roomId={roomTimeline.roomId} />}
+      </div>
```
ajbura: We don't need to wrap PeopleDrawer in a div as we can directly apply media query on `.people-drawer` class. currently wrapping that into a div affect the people-drawer layout from taking up all the vertical space.

src/app/organisms/room/Room.scss
```diff
@@ -9,4 +10,12 @@
...
     overflow: hidden;
   }
-}
\ No newline at end of file
+}
+
+.room__drawer {
+  @include threshold.smaller(tablet) {
+    // On compact screen sizes there is not enough space
+    // but no need to view it otherwise as it already is in settings
```
ajbura: please don't use comments in css. mixin and display: none; already tell the story.

src/app/organisms/room/RoomViewHeader.jsx
```diff
@@ -73,6 +74,13 @@ function RoomViewHeader({ roomId }) {
 
   return (
     <Header>
+      <div className="room-back-btn">
+        <IconButton
+          src={BackArrowIC}
+          tooltip="Return to navigation"
+          onClick={() => navigation.emit(cons.events.navigation.OPEN_NAVIGATION)}
+        />
+      </div>
```
ajbura: Please use BEM methodology for classNames. also we should just pass classname to IconButton component itself so we don't need to wrap that inside div.

src/app/organisms/room/RoomViewHeader.jsx
```diff
@@ -87,7 +95,9 @@ function RoomViewHeader({ roomId }) {
         <RawIcon src={ChevronBottomIC} />
       </button>
       <IconButton onClick={() => toggleRoomSettings(tabText.SEARCH)} tooltip="Search" src={SearchIC} />
-      <IconButton onClick={togglePeopleDrawer} tooltip="People" src={UserIC} />
+      <div className="room-people-drawer-toggle">
+        <IconButton onClick={togglePeopleDrawer} tooltip="People" src={UserIC} />
+      </div>
```
ajbura: same as above.

src/app/partials/threshold.scss
```diff
@@ -0,0 +1,28 @@
+
+$tablet-threshold: 900px;
+$mobile-threshold: 750px;
```
ajbura: I thing we should rename these variable to $breakpoint-mobile and $breakpoint-tablet and filename as screen so things make more sense when we use these mixins like screen.mixinName instead of threshold.mixinName.

src/app/partials/threshold.scss
```diff
@@ -0,0 +1,28 @@
...
+@mixin bigger($devicetype) {
+  @if $devicetype==mobile {
+    @media screen and (min-width: $mobile-threshold) {
+      @content;
+    }
+  } @else if $devicetype==tablet {
+    @media screen and (min-width: $tablet-threshold) {
+      @content;
+    }
+  }
```
ajbura: it will be more readable if we call these mixin's as `smallerThan` and `biggerThan` as they function like that. Also `deviceType` doesn't fit with the intent here we may call that `deviceBreakpoint` as it's value doesn't define a device but a point form where one side is let's say mobile and other is tablet. And they will be used like `screen.smallerThan(mobileBreakpoint)` and `screen.biggerThan(tabletBreakpoint)` which sounds more descriptive than current one.

src/app/templates/client/Client.jsx
```diff
@@ -16,11 +16,33 @@ import navigation from '../../../client/state/navigation';
...
+    navigation.on(cons.events.navigation.OPEN_NAVIGATION, onNavigationSelected);
+    // appDispatcher.register(cons.events.navigation.ROOM_SELECTED, () => setOnNavRoom(false));
+
+    return (() => {
+      navigation.removeListener(onRoomSelected);
+    });
+  }, []);
+  // #endregion
 
+  // #region Startup
```
ajbura: Please remove unnecessary comments.

src/app/templates/client/Client.jsx
```diff
@@ -16,11 +16,33 @@ import navigation from '../../../client/state/navigation';
...
+  const [activeView, setActiveView] = useState(viewPossibilities.nav);
+
+  // #region Liston on events for compact screen sizes
+  const onRoomSelected = () => setActiveView(viewPossibilities.room);
+  const onNavigationSelected = () => setActiveView(viewPossibilities.nav);
+
+  useEffect(() => {
+    navigation.on(cons.events.navigation.ROOM_SELECTED, onRoomSelected);
+    navigation.on(cons.events.navigation.OPEN_NAVIGATION, onNavigationSelected);
+    // appDispatcher.register(cons.events.navigation.ROOM_SELECTED, () => setOnNavRoom(false));
```
ajbura: please remove unused code.

src/app/templates/client/Client.jsx
```diff
@@ -16,11 +16,33 @@ import navigation from '../../../client/state/navigation';
...
+
+    return (() => {
+      navigation.removeListener(onRoomSelected);
+    });
```
ajbura: listeners are not being removed properly see: https://nodejs.org/dist/v11.13.0/docs/api/events.html#events_emitter_removelistener_eventname_listener

src/client/state/cons.js
```diff
@@ -91,6 +92,7 @@ const cons = {
       REPLY_TO_CLICKED: 'REPLY_TO_CLICKED',
       SEARCH_OPENED: 'SEARCH_OPENED',
       REUSABLE_CONTEXT_MENU_OPENED: 'REUSABLE_CONTEXT_MENU_OPENED',
+      OPEN_NAVIGATION: 'OPEN_NAVIGATION',
```
ajbura: we should rename this to NAVIGATION_OPENED cuz it's an event that happened and we are notifying about it to UI.

src/app/templates/client/Client.jsx
```diff
@@ -123,10 +148,14 @@ function Client() {
       onDragLeave={handleDragLeave}
       onDrop={handleDrop}
     >
-      <div className="navigation__wrapper">
+      <div className={`navigation__wrapper
+        ${activeView !== viewPossibilities.nav ? 'non-focused-ui' : ''}`}
```
ajbura: `non-focused-ui` classname doesn't follow BEM methodology and can cuz class name collision.

src/app/organisms/room/RoomViewHeader.scss
```diff
@@ -24,4 +25,19 @@
...
+
+.room-people-drawer-toggle {
+  @include threshold.smaller(tablet) {
+    display: none;
+  }
+}
+
+.room-back-btn {
+  margin-left: -10px;
+  margin-right: 5px;
```
ajbura: please don't use negative margins. also we uses spacing variables(already defined in index.scss) instead of direct values. also use dir.side mixin for side margin/padding

nitanmarcel:
Using this PR as a base for an ubuntu touch app. Nicely done.

One thing that's missing is the 3-dots menu on the room list views which shows outside the view.

Also touch doesn't send an hover event which bugs the message context menu (reply, edit) so probably that needs to be handled in mobile view too somehow.

Looking forward on this PR :)

image

src/app/organisms/room/RoomViewHeader.jsx
```diff
@@ -74,13 +74,12 @@ function RoomViewHeader({ roomId }) {
...
-      <div className="room-back-btn">
-        <IconButton
-          src={BackArrowIC}
-          tooltip="Return to navigation"
-          onClick={() => navigation.emit(cons.events.navigation.OPEN_NAVIGATION)}
-        />
-      </div>
+      <IconButton
+        src={BackArrowIC}
+        className="room_back__button"
```
ajbura: className doesn't follow BEM. it will be like `room-header__back-btn`

src/app/organisms/room/RoomViewHeader.jsx
```diff
@@ -95,9 +94,7 @@ function RoomViewHeader({ roomId }) {
         <RawIcon src={ChevronBottomIC} />
       </button>
       <IconButton onClick={() => toggleRoomSettings(tabText.SEARCH)} tooltip="Search" src={SearchIC} />
-      <div className="room-people-drawer-toggle">
-        <IconButton onClick={togglePeopleDrawer} tooltip="People" src={UserIC} />
-      </div>
+      <IconButton className="room-people-drawer__toggle" onClick={togglePeopleDrawer} tooltip="People" src={UserIC} />
```
ajbura: className doesn't follow BEM. it will be like `room-header__drawer-btn`. ([block]__[element]--[modifier]) block is room-header, element is drawer-btn

src/app/templates/client/Client.jsx
```diff
@@ -149,12 +146,12 @@ function Client() {
...
     >
       <div className={`navigation__wrapper
-        ${activeView !== viewPossibilities.nav ? 'non-focused-ui' : ''}`}
+        ${activeView !== viewPossibilities.nav ? 'client-container-item__hidden' : null}`}
       >
         <Navigation />
       </div>
       <div className={`room__wrapper
-        ${activeView !== viewPossibilities.room ? 'non-focused-ui' : ''}`}
+        ${activeView !== viewPossibilities.room ? 'client-container-item__hidden' : null}`}
```
ajbura: className doesn't follow BEM methodology.

src/client/state/cons.js
```diff
@@ -47,7 +47,7 @@ const cons = {
       CLICK_REPLY_TO: 'CLICK_REPLY_TO',
       OPEN_SEARCH: 'OPEN_SEARCH',
       OPEN_REUSABLE_CONTEXT_MENU: 'OPEN_REUSABLE_CONTEXT_MENU',
-      OPEN_NAVIGATION: 'OPEN_NAVIGATION',
+      NAVIGATION_OPENED: 'NAVIGATION_OPENED',
```
ajbura: This can be as it was before (OPEN_NAVIGATION) because it is an action that user calls. Other one was event (already has happened). It's like user say OPEN_NAVIGATION and system open it and inform view NAVIGATION_OPENED.

src/app/templates/client/Client.jsx
```diff
@@ -33,16 +33,14 @@ function Client() {
...
     navigation.on(cons.events.navigation.ROOM_SELECTED, onRoomSelected);
-    navigation.on(cons.events.navigation.OPEN_NAVIGATION, onNavigationSelected);
-    // appDispatcher.register(cons.events.navigation.ROOM_SELECTED, () => setOnNavRoom(false));
+    navigation.on(cons.events.navigation.NAVIGATION_OPENED, onNavigationSelected);
 
     return (() => {
-      navigation.removeListener(onRoomSelected);
+      navigation.removeListener(cons.events.navigation.NAVIGATION_OPENED, onNavigationSelected);
     });
   }, []);
```
ajbura: ROOM_SELECTED listener is not being removed.

src/app/templates/client/Client.jsx
```diff
@@ -123,10 +145,14 @@ function Client() {
...
-      <div className="navigation__wrapper">
+      <div className={`navigation__wrapper
+        ${activeView !== viewPossibilities.nav ? 'client-container-item__hidden' : null}`}
+      >
         <Navigation />
       </div>
-      <div className="room__wrapper">
+      <div className={`room__wrapper
+        ${activeView !== viewPossibilities.room ? 'client-container-item__hidden' : null}`}
+      >
```
ajbura: Unnecessary line-breaks are added to the className. Also null is added to the className if condition is false. 

I think using `ref` to change their visibility will be more good instead of re-rendering this component which is very costly. Also with ref we can also add nice animation to make it feel more smooth.

C0ffeeCode:
Which BEM classNames would you propose?

src/app/templates/client/Client.jsx
ajbura: className `client__item-hidden`

ajbura:
Commented above in review comments.

thebiblelover7:
Looking forward for this PR to be merged! Would love to use cinny on my phone!

ajbura (approved):
Amazing work. Thanks :)

nitanmarcel:
Awesome. I can base my UT app on the upstream now.

Thanks for merging this

PR #499, opened 2022-04-24
Show devices without encryption support separately from unverified session

ginnyTheCat:
### Description

Cinny marks devices without encryption support as not verified. They however are unable to be verified. The warning is therefore not very useful (as there is no way to fix it, and the device not having access to encrypted messages to begin with).

Element Web also does this
image

This also makes these devices not appear in the "unverified sessions" warning in the navigation bar.

Before:
image

After:
image

#### Type of change

### Checklist:

Preview: https://62657afdcc0e5743464f5e2c--pr-cinny.netlify.app

### DIFF
--- a/src/app/organisms/navigation/SideBar.jsx
+++ b/src/app/organisms/navigation/SideBar.jsx
@@ -92,7 +92,7 @@ function ProfileAvatarMenu() {
 
 function CrossSigninAlert() {
   const deviceList = useDeviceList();
-  const unverified = deviceList?.filter((device) => !isCrossVerified(device.device_id));
+  const unverified = deviceList?.filter((device) => isCrossVerified(device.device_id) === false);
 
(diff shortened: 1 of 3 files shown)


ajbura:
we can also remove the red text color for them

ginnyTheCat:
I like the red text as it shows that it's still not really as secure as it should/could be.
Still I wouldn't mind changing it.

ajbura:
I choose the red color so it prompt user to verify the session. but if we can't verify it, it will create disturbance.

Apart from this we still gonna have that (unverified session)alert on sidebar for can't verify sessions? EDITED: ok i see it 👍

ginnyTheCat:
>Apart from this we still gonna have that (unverified session)alert on sidebar for can't verify sessions?

No, that was the main reason I created this fix.

ajbura (approved):

PR #564, opened 2022-05-19
Add typing notification and read receipt toggle

tezlm:
adds a typing notification and read receipt toggle. not sure if the placement (currently appearance) is a good place to put it though


kfiven:
For the read receipt part there's an MSC https://github.com/matrix-org/matrix-spec-proposals/pull/2285

kfiven:
Closed as stale.

PR #714, opened 2022-08-04
Show full timestamp on hover

### LINKED ISSUE #684: View specific timestamps of older messages
### Describe the problem

I don't see any way to view the specific timestamp (time and not just date) of messages sent yesterday or earlier. Instead, those messages only display the date. This can be frustrating especially at midnight when all the most recent messages suddenly lose their timestamp in the UI.

I would like to see the time and date of the message for example in a tooltip when hovering over the date displayed in the right edge.

Additionally, maybe consider displaying the time instead of date for all messages of the past 24 hours, rather than messages sent today?

### Alternatives considered

_No response_

### Additional context

_No response_

ginnyTheCat:
### Description
Shows the full date when hovering over a time of a message or event. It also changes it to not be date anymore for past days as the date can now be found out by hovering over the time.

image

Fixes #684

#### Type of change

### Checklist:

Preview: https://62ed22314d1e7c6b09c6ffc4--pr-cinny.netlify.app

### DIFF
--- a/src/app/atoms/time/Time.jsx
+++ b/src/app/atoms/time/Time.jsx
@@ -4,29 +4,41 @@ import PropTypes from 'prop-types';
 import dateFormat from 'dateformat';
 import { isInSameDay } from '../../../util/common';
 
-function Time({ timestamp }) {
+function Time({ timestamp, fullTime }) {
   const date = new Date(timestamp);
 
-  const compareDate = new Date();
-  const isToday = isInSameDay(date, compareDate);
+  const formattedFullTime = dateFormat(date, 'dd mmmm yyyy, hh:MM TT');
+  let formattedDate = formattedFullTime;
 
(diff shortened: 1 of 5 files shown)


kfiven:
The format we planned when this was added was:
* HH:MM AM/PM for the current day.
* Yesterday HH:MM AM/PM for the past one day. (wasn't added)
* DD/MM/YYYY for the rest of the timeline.
And upon hovering over all of these we can see the full timestamp.

Since the date separator scroll with timeline, the HH:MM in all cases gets really confusing and was the reason for date for older messages.

ginnyTheCat:
Should I add the yesterday case in as well?

kfiven:
That'll be good addition.

ajbura (commented):
Message component is also used in RoomSearch.jsx so we need to change the time prop to timestamp there also. 

Otherwise looks good

ajbura (approved):

PR #729, opened 2022-08-09
Update prod-deploy.yml

GoliathLabs:
### Description

I've added the GH-Cache setting for the build-push action. This should speed up the build process.

Fixes #

#### Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings

Preview: https://62f2df6c8a2c4f25bef1bdcc--pr-cinny.netlify.app
⚠️ Exercise caution. Use test accounts. ⚠️

PR #730, opened 2022-08-09
Added two recommended flags & bumped image tag

GoliathLabs:
### Description

I also introduced the flags `--no-audit` and `--prefer-offline`. We don't need an audit for the packages we use, because we already use Dependabot to alert us on repository level. With `--prefer-offline` we try to use the already downloaded packages (which is very useful when #729 is merged).

Fixes #

#### Type of change

### Checklist:

- [X] I have made corresponding changes to the documentation

Preview: https://62f2df41eaa7052c9bc60b85--pr-cinny.netlify.app

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 ## Builder
-FROM node:17.9.1-alpine3.15 as builder
+FROM node:17.9.0-alpine3.15 as builder
 
 WORKDIR /src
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,10 @@
 ## Builder
-FROM node:17.9.0-alpine3.15 as builder
+FROM node:17.9.1-alpine3.15 as builder
```
kfiven: Don't bump node. There are issues with npm >8.5.5.

kfiven (changes requested):

PR #835, opened 2022-09-10
Close settings and search with escape key

MyBeansAreBaked:
### Description

Before, the search dialogue could only be dismissed using the mouse.
The settings page could sometimes be dismissed with escape, unless you had interacted with it.

This change allows you to dismiss the search or settings page just by pressing escape.
Also, the search dialogue can be dismissed by pressing backspace on an empty search.

This change adds a new event REQUEST_CLOSE, which is triggered by pressing escape, this could be implemented by other dialogues too.
It also fixes what looked to be a mistake that the isRawModalVisible setter, wasn't a proper setter.

#### Type of change

### Checklist:

Preview: https://835--pr-cinny.netlify.app


kfiven:
Hi, thanks for the pull request. But both settings and search already close with `Esc` key and I am unsure why you tried to reimplement that.

MyBeansAreBaked:
On 22/09/10 04:51AM, Krishan wrote:
> both settings and search already close with `Esc` key and I am unsure why you tried to reimplement that.

Ah, I got ahead of myself. Sorry for not investigating further.
I see that the search dialogue does close after pressing escape in
Firefox and the Tauri build, but in the Brave browser this is not the
case.

In all browsers, the settings page does close if you press escape
immediately after opening it. However this seems to not be explicitly
defined anywhere, and *after moving your focus into the dialogue* (just
click on anything inside of it), pressing escape will not close it.

This change will make the behavior more consistent.

PR #847, opened 2022-09-14
Parsing HTML to Markdown AST

ginnyTheCat:
### Description
This parses the `formatted_body` (if available) into the markdown ast to then render it as plaintext. This allows spoilers to not be visible in notifications (now showing up as `<spoiler>`), while still showing during edits as `||spoiler||` and in the future as `[spoiler](mxc://somewhere)` in the `body` shown in plaintext clients.

#### Type of change 

### Checklist:

Preview: https://847--pr-cinny.netlify.app

### DIFF
--- a/src/app/molecules/message/Message.jsx
+++ b/src/app/molecules/message/Message.jsx
@@ -803,7 +803,9 @@ function Message({
         )}
         {isEdit && (
           <MessageEdit
-            body={(customHTML ? html : plain)(customHTML || body, { kind: 'edit' }).plain}
+            body={(customHTML
+              ? html(customHTML, { kind: 'edit', onlyPlain: true }).plain
+              : plain(body, { kind: 'edit', onlyPlain: true }).plain)}
             onSave={(newBody) => {
               if (newBody !== body) {
                 initMatrix.roomsInput.sendEditedMessage(roomId, mEvent, newBody);
--- a/src/app/molecules/message/Message.jsx
+++ b/src/app/molecules/message/Message.jsx
@@ -40,6 +40,7 @@ import BinIC from '../../../../public/res/ic/outlined/bin.svg';
 
 import { confirmDialog } from '../confirm-dialog/ConfirmDialog';
 import { getBlobSafeMimeType } from '../../../util/mimetypes';
+import { html, plain } from '../../../util/markdown';
 
 function PlaceholderMessage() {
   return (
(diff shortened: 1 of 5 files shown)


src/app/molecules/message/Message.jsx
```diff
@@ -802,7 +803,7 @@ function Message({
         )}
         {isEdit && (
           <MessageEdit
-            body={body}
+            body={(customHTML ? html : plain)(customHTML || body, { kind: 'edit' }).plain}
```
ajbura: This condition feels confusing. we can just simply do `customHTML ? html(....) : plain( .... )`

ajbura: This condition feels confusing. we can just simply do `customHTML ? html(....) : plain( .... )`

src/util/common.js
```diff
@@ -204,3 +204,27 @@ export function scaleDownImage(imageFile, width, height) {
...
+ * @param {flags} string regex flags
+ * @param {prefix} string prefix appended at the beginning of the regex
+ * @returns {RegExp}
+ */
+export function idRegex(sigil, flags, prefix) {
+  const servername = '(?:[a-zA-Z0-9-.]*[a-zA-Z0-9]+|\\[\\S+?\\])(?::\\d+)?';
+  return new RegExp(`${prefix}(${sigil}\\S+:${servername})`, flags);
+}
+
+const matrixToRegex = /^https?:\/\/matrix.to\/#\/(.+:.+)/;
```
ajbura: at the end we have `(.+:.+)` we can convert it to `\S+:\S+` so it doesn't match white spaces as mxid's don't have them.

src/util/common.js
```diff
@@ -204,3 +204,27 @@ export function scaleDownImage(imageFile, width, height) {
...
+ * Parses a matrix.to URL into an matrix id.
+ * This function can later be extended to support matrix: URIs
+ * @param {string} uri The URI to parse
+ * @returns {string} The id or null if the URI does not match
+ */
+export function parseIdUri(uri) {
+  const res = decodeURIComponent(uri).match(matrixToRegex);
+  if (!res) return null;
+  return res[1];
+}
```
ajbura: This function also return null but jsdoc only talk about string.

ginnyTheCat: It says it in the `@returns` part but i can add it to the description as well

ajbura: I mean `@returns {string}` should be `@returns {string|null}` so IDE can give right suggestion

ajbura (approved):
Great work. Thank you!

PR #1152, opened 2023-03-09
Fix messages getting sent when confirming an input method selection

soundflora:
### Description

When using a CJK input method, the enter key is used to confirm the current conversion, so it shouldn't send the message yet. Normally this can be checked with the `isComposing` property of the native event, but that doesn't work on the macOS webview.

Instead, check for `keyCode` 229, which is what you get when confirming conversions, and ignore it.

#### Type of change

### Checklist:


kfiven:
Closing as stale.

PR #1207, opened 2023-04-19
Simplify space to room mapping logic

### LINKED ISSUE #353: Clicking a side bar 'tab' should bring you to the last chat you were looking at on that tab.
Whenever I have to switch spaces/to the home tab/to the dm tab, I feel like I'm constantly 'catching up' to where I mentally already am. I'm almost never going to switch tabs, while still wanting to look at the chat I was already looking at, so having to constantly click onto the new tab, and then find and click on the chat in that tab I want to view, can be tiresome.

Clicking a side bar 'tab/space/etc.' should save & bring you back to the last chat you were looking at on that tab. (how discord & element do it)

#### Additional context
this is the "tab" im talking about
image

greentore:
### Description
Simplifies logic controlling which room will be displayed when you select a space. Right now it's pretty complicated, but with this patch it will always be the last room that was selected when the space was selected, which I believe is the most intuitive behavior.

Fixes #353 a bit more

#### Type of change

### Checklist:

### DIFF
--- a/src/client/state/navigation.js
+++ b/src/client/state/navigation.js
@@ -39,7 +39,7 @@ class Navigation extends EventEmitter {
   }
 
   _mapRoomToSpace(roomId) {
-    const { roomList, accountData } = this.initMatrix;
+    const { roomList } = this.initMatrix;
     if (
       this.selectedTab === cons.tabs.HOME
       && roomList.rooms.has(roomId)
@@ -59,23 +59,10 @@ class Navigation extends EventEmitter {
       return;
     }
 
-    const parents = roomList.roomIdToParents.get(roomId);
-    if (!parents) return;
-    if (parents.has(this.selectedSpaceId)) {
-      this.spaceToRoom.set(this.selectedSpaceId, {
-        roomId,
-        timestamp: Date.now(),
-      });
-    } else if (accountData.categorizedSpaces.has(this.selectedSpaceId)) {
-      const categories = roomList.getCategorizedSpaces([this.selectedSpaceId]);
-      const parent = [...parents].find((pId) => categories.has(pId));
-      if (parent) {
-        this.spaceToRoom.set(parent, {
-          roomId,
-          timestamp: Date.now(),
-        });
-      }
-    }
+    this.spaceToRoom.set(this.selectedSpaceId, {
+      roomId,
+      timestamp: Date.now(),
+    });
   }
 
   _selectRoom(roomId, eventId) {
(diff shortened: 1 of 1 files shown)


src/client/state/navigation.js
```diff
@@ -59,23 +59,10 @@ class Navigation extends EventEmitter {
...
-    } else if (accountData.categorizedSpaces.has(this.selectedSpaceId)) {
-      const categories = roomList.getCategorizedSpaces([this.selectedSpaceId]);
-      const parent = [...parents].find((pId) => categories.has(pId));
-      if (parent) {
-        this.spaceToRoom.set(parent, {
-          roomId,
-          timestamp: Date.now(),
-        });
-      }
-    }
```
ajbura, line 78: What this portion does is it check parent parent for room in when current selected space is mark as categorized. I looks to find the possible parent in those spaces(aka spaces that are displayed as categories). 

I don't know why we want to remove this behaviour as it will break it.

src/client/state/navigation.js
```diff
@@ -212,13 +199,6 @@ class Navigation extends EventEmitter {
     if (categorizedSpaces.has(spaceId)) {
       const categories = roomList.getCategorizedSpaces([spaceId]);
 
-      const latestSelectedRoom = this._getLatestSelectedRoomId([...categories.keys()]);
-
-      if (latestSelectedRoom) {
-        this._selectRoom(latestSelectedRoom);
-        return;
-      }
-
```
ajbura, line 221: This code get the last selected room when a selected space is marked as categorized. It takes all those spaces as categories and finds room which was most recently selected.

ajbura (commented):
currently we do select last selected room but on initial application load when we haven't selected any room in space we select one which has most recent message.

src/client/state/navigation.js
greentore: The issue with this logic is that when there's a shared space nested in 2 separate spaces selecting a room from the shared space in one will mark it as last selected in the other one as well. I find this behavior really undesirable.

Consider the following layout:
Space `A` containing room `a` and space `S`
Space `B` containing room `b` and space `S`
Space `S` containing room `s`

Let's assume that spaces `A` and `B` are pinned to the sidebar with categorized spaces enabled. You click on space `A` and select room `a`, then you open space `B` and select nested room `s`. Finally you go back to space `A` and instead of expected room `a` you see room `s`.

I know my patch is really half-assed, but I have been running with it for several months with no issues so far.

greentore:
v4 fixed my issues, so this patch is no longer needed.

PR #1346, opened 2023-07-21
Refactor timeline

ajbura:
### Description

Fixes #

#### Type of change

### Checklist:


tezlm:
Some feedback after test driving the new ui design, sorry in advance if this is a lot or planned to be added/fixed

- The message right click/context menu is nice, but it would be nicer if it opened the menu where the click happened
- The buttons seem harder to click or don't "feel as good", the padding should probably be increased (while maybe decreasing margins)
	- The message toolbar seems finnicky to access, since i can't approach it an angle (only directly from the bottom)
	- The "back button" on the file preview modal is tiny and seeminly useless since I can click on the background
- The design being partially old style and partially new style/folds is jarring at times
- Sometimes uploading files can upload as the wrong type or octet-stream (ie. typescript, mov occasionally)
- There's no way to edit a message currently
- There's no way to hide deleted message tombstones
- The image viewer lets me pan the image by dragging, but only when it's not at 100% zoom; i don't know if its better to be able to drag/drop images or have a consistent ui
- Media no longer has a "one click download" button, which looks cleaner but does hurt usability a bit (i like to archive stuff)
- Showing the user id separately from the username (instead of hover) is and (imo) better overall, but might clutter the ui a bit
- Typing notifications don't seem to currently work (note: this might be my server having issues)
- There's no indication that a `<details>` element is clickable, maybe add `summary { cursor: pointer; }`?
- Tables don't have any styling
- `m.notice`s changing color is nice but currently seems a bit too subtle. `m.emote` stands out too much
- `<code class="language-*">` outside of `<pre>`s have broken styling
- Clicking "watch" on a video can cause the browser spinner and the cinny spinner to load at the same time
- LaTeX no longer works
- "View Reactions" is prone to breaking
- There is no indicator for encrypted rooms
- Opening the reaction picker seems a bit laggy, try memoizing it if you aren't already
- There is no indicator for messages that notified me (i can see mentions, but not reply or keyword pings)
- Location messages used to show previews (in previous commits) but seems to have been broken?
- Maybe the text preview modal should render markdown files? If this is added, a toggle could be added to switch between plain/rendered views.
- The message toolbar icons might not disappear if you right click multiple messages at a time ([screenshot](https://github.com/cinnyapp/cinny/assets/56274831/f065868a-8644-44bf-aa10-5ff21bd52428))

Despite all this, the redesign is really nice! Thank you for working on Cinny

ajbura:
> "View Reactions" is prone to breaking

@tezlm can you provide more details on this?

tezlm:
This message supposedly has 6 bird reactions:

image

But viewing reactions only shows 5:

image

This was loaded by scrolling up, but it can also happen from search results, jump to unread, jump to reply, etc.

ajbura:
ok thanks. 

I am going to merge this now. And will add missing features or fix bug in next one.

PR #1349, opened 2023-07-23
Add basic `m.thread` support

### LINKED ISSUE #257: Threads

It can get very cluttered with everyone talking in the same channel. It would be really, really amazing to have threads like in discord. They're kind of like sub-channels that are tied to a specific message and can be created by anyone. For instance, someone could bring up an issue and we could create a thread on that message.

Proof that it is possible:
https://github.com/matrix-org/cerulean

greentore:
### Description
Adds basic `m.thread` support as described in [MSC3440](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#sending-mthread-before-fully-implementing-threads).
Support in this case means that replying to a threaded message will place the reply in the thread in clients that support them.

Partially fixes #257

#### Type of change

### Checklist:


src/app/organisms/room/RoomInput.tsx
```diff
@@ -284,6 +286,10 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
             event_id: replyDraft.eventId,
           },
         };
+        if (replyDraft.relatesTo?.rel_type === "m.thread") {
+          content['m.relates_to'].event_id = replyDraft.relatesTo.event_id;
+          content['m.relates_to'].rel_type = "m.thread";
+        }
```
ajbura: As we do not support thread in cinny we should also add `"is_falling_back": true` as we are replying to a message which is part of thread compare to sending a message in thread with client that support them.

src/app/state/roomInputDrafts.ts
```diff
@@ -39,7 +40,8 @@ export type IReplyDraft = {
   userId: string;
   eventId: string;
   body: string;
-  formattedBody?: string;
+  formattedBody: string | null;
+  relatesTo: IEventRelation | undefined;
```
ajbura: Not sure why `formattedBody?: string` is changed to `string | null`. `relatesTo` should also be optional?

src/app/organisms/room/RoomInput.tsx
```diff
@@ -185,13 +185,15 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
         userId: string,
         eventId: string,
         body: string,
-        formattedBody: string
+        formattedBody: string | null,
+        relatesTo: IEventRelation | undefined,
```
ajbura: Should not we make them optional as well?

src/app/organisms/room/RoomInput.tsx
ajbura: https://spec.matrix.org/v1.6/client-server-api/#replies-within-threads spec for ref

src/app/state/roomInputDrafts.ts
greentore: I added `| null` as Typescript would complain about it with `exactOptionalPropertyTypes` enabled.  You're right about optionality, it escaped my mind.

src/app/organisms/room/RoomInput.tsx
greentore: It's the other way around. If we sent `"is_falling_back": true`, the message would be interpreted as posted to the thread directly with `m.in_reply_to` being set only for fallback purposes.

naskya:
any updates?

filipesmedeiros:
Is any help needed here? I can try to contribute if you think I could help :)

JuliaSoriaSmith:
Can the reviewers with write access please look at this? This has been open for quite some time and threads are a noticeable missing feature

src/app/organisms/room/RoomInput.tsx
greentore: This time I explicitly set `is_falling_back` to `false`. I think it's saner considering we're making a normal reply from Cinny's perspective.

greentore:
I added threaded reply indicators inspired by element-x. Should resolve https://github.com/cinnyapp/cinny/issues/1871.

src/app/components/message/Reply.css.ts
```diff
@@ -5,6 +5,23 @@ export const ReplyBend = style({
...
+  selectors: {
+    ':hover&': {
+      opacity: config.opacity.P500,
+    },
+  },
+});
+
+export const ThreadIndicatorIcon = style({
+  width: '0.875rem',
+  height: '0.875rem',
```
greentore: I tried using `size` and `gap` props from `folds`, but the smallest size was still too big for this element.

williamkray:
i'm not an approver but i've briefly tested the netlify deployment of this PR and it looks fantastic, i'm very eager to see this merged.

seems like the quoted message doesn't get truncated with elipses (...) properly, and runs off the right hand side of the screen.

image

williamkray:
same messages in stable cinny:

image

greentore:
@williamkray looks like I broke it again. Thanks for the report.

williamkray:
looks good after the fix!

src/app/components/message/Reply.css.ts
ajbura: Please use `toRem` function from folds to write `rem` values in `px` as `rem` unit are hard to coordinate with designs.

ajbura (approved):
Thanks for the PR, it looks great to me.

Could you please mark the PR as ready for review.

greentore:
I converted rem units to equivalent `toRem` calls as requested.
I also considered disabling thread support in sdk, but resulting unthreaded read receipts mark all threads in the room as read in threaded clients, which may be considered undesired behavior.

PR #1359, opened 2023-07-30
Prevent `manifest.json` from being inlined

greentore:
### Description
With vite's asset inlining enabled `manifest.json` isn't detected, which results in browsers not treating Cinny as an installable webapp.

Fixes #

#### Type of change

### Checklist:

### DIFF
--- a/index.html
+++ b/index.html
@@ -27,7 +27,7 @@
 
     <link id="favicon" rel="shortcut icon" href="./public/favicon.ico" />
 
-    <link rel="manifest" href="./public/manifest.json" />
+    <link rel="manifest" href="/manifest.json" />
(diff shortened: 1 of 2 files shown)


ajbura:
Inline asset was enabled because we load system icons as -webkit-mask-image in css property and without enabling inline asset it take noticeable delay to load those icons which is not good in term of ux. So I think it won't be a good idea to disable it until we implement folds design system in cinny.

greentore:
I found a way to prevent `manifest.json` from being inlined without disabling inlining entirely.

ajbura (approved):
Looks good. Thanks for fixing it!

PR #1448, opened 2023-10-13
add sanitizing out the nested mx-reply

jjj333-p:
### Description

Fixes # removes the nested mxreply in the message you're quoting so that it doesnt make it apart of message body. this is due to the reply removing the nested `<mx-reply>` but not the `</mx-reply>` causing it to end the html tag early
image
image

probably need to change it to make it typescript, i only know js. im praying that this works fine just popping in js.

#### Type of change

### Checklist:


jjj333-p:

PR #1555, opened 2023-11-21
make small attachments not scale in timeline

WebFreak001:
### Description

ultra wide / ultra tall but otherwise small images are cropped if too small in one dimension (minimum size changed to 32x32 in timeline for images, 48x48 for videos)

preview:

example timeline

#### Type of change

- [x] Design change

### Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas - n/a

### DIFF
--- a/src/app/organisms/room/msgContent.ts
+++ b/src/app/organisms/room/msgContent.ts
@@ -53,7 +53,7 @@ export const getImageMsgContent = async (
     body: file.name,
   };
   if (imgEl) {
-    const dim = scaleDimension(imgEl.width, imgEl.height, 512, 32, 512, 32);
+    const dim = scaleDimension(imgEl.width, imgEl.height, 32, 32, 512, 512);
     const blurHash = encodeBlurHash(imgEl, dim.w, dim.h);
 
     content.info = {
@@ -96,7 +96,7 @@ export const getVideoMsgContent = async (
       )
     );
     if (thumbContent && thumbContent.thumbnail_info) {
-      const dim = scaleDimension(videoEl.width, videoEl.height, 512, 32, 512, 32);
+      const dim = scaleDimension(videoEl.width, videoEl.height, 32, 32, 512, 512);
       thumbContent.thumbnail_info[MATRIX_BLUR_HASH_PROPERTY_NAME] = encodeBlurHash(
         videoEl,
         dim.w,
--- a/src/app/organisms/room/msgContent.ts
+++ b/src/app/organisms/room/msgContent.ts
@@ -12,7 +12,7 @@ import {
 import { encryptFile, getImageInfo, getThumbnailContent, getVideoInfo } from '../../utils/matrix';
 import { TUploadItem } from '../../state/roomInputDrafts';
 import { encodeBlurHash } from '../../utils/blurHash';
-import { scaleYDimension } from '../../utils/common';
+import { scaleDimension } from '../../utils/common';
 
 const generateThumbnailContent = async (
   mx: MatrixClient,
@@ -53,7 +53,8 @@ export const getImageMsgContent = async (
     body: file.name,
   };
   if (imgEl) {
-    const blurHash = encodeBlurHash(imgEl, 512, scaleYDimension(imgEl.width, 512, imgEl.height));
+    const dim = scaleDimension(imgEl.width, imgEl.height, 512, 32, 512, 32);
+    const blurHash = encodeBlurHash(imgEl, dim.w, dim.h);
 
     content.info = {
       ...getImageInfo(imgEl, file),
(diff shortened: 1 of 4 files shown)


WebFreak001:

src/app/organisms/room/msgContent.ts
```diff
@@ -95,10 +96,11 @@ export const getVideoMsgContent = async (
...
     );
     if (thumbContent && thumbContent.thumbnail_info) {
+      const dim = scaleDimension(videoEl.width, videoEl.height, 512, 32, 512, 32);
       thumbContent.thumbnail_info[MATRIX_BLUR_HASH_PROPERTY_NAME] = encodeBlurHash(
         videoEl,
-        512,
-        scaleYDimension(videoEl.videoWidth, 512, videoEl.videoHeight)
+        dim.w,
+        dim.h,
       );
```
ajbura, line 104: These changes doesn't add any new functionality?

src/app/utils/common.ts
```diff
@@ -63,10 +63,17 @@ export const binarySearch = <T>(items: T[], match: (item: T) => -1 | 0 | 1): T |
 export const randomNumberBetween = (min: number, max: number) =>
   Math.floor(Math.random() * (max - min + 1)) + min;
 
-export const scaleYDimension = (x: number, scaledX: number, y: number): number => {
-  const scaleFactor = scaledX / x;
-  return scaleFactor * y;
-};
+export const scaleDimension = (x: number, y: number, maxX: number, minX: number = maxX, maxY: number = Infinity, minY: number = 0): {w: number, h: number} => {
```
ajbura: I think for better readability `minX` and `minY` should comes earlier than `maxX` and `maxY`?
eg: `(x, y, minX, minY, maxX, maxY)`

src/app/utils/common.ts
```diff
@@ -63,10 +63,17 @@ export const binarySearch = <T>(items: T[], match: (item: T) => -1 | 0 | 1): T |
...
+  let w = Math.max(minX, Math.min(maxX, x));
+  const scaleFactor = w / x;
+  let h = Math.max(minY, scaleFactor * y);
+  if (h > maxY) {
+    w = w / h * maxY;
+    h = maxY;
+  }
+  w = Math.max(minX, Math.min(maxX, w));
+  return { w, h };
```
ajbura, line 75: This function took dimension in term of `x` and `y` so returning them as `w` and `h` on object make them appear differently on output, either we should return it as tuple `[scaledX, scaledY]` or call them `const scaled = { x, y }`.

ajbura (changes requested):
These changes brings following issues:
1. By collapsing image width we won't be able to distinguish between what is sticker and what is image.
2. Making sticker width between 48 to 152 means adding inconsistence UI for sticker with different width, they should follow same size just like custom emoji's.
3. And if image/sticker width is really small it clutter up the `View` image button and `image info` when Auto Load Media feature is disabled.

We can ignore first issue as transparent image has tinted background but transparent sticker does not have that. For second we can revert the changes as it doesn't feel like a good idea to render sticker based on image info width. And for last we can increase in minimum width/height to at least 128

src/app/utils/common.ts
ajbura: Now I got it. We should also make them all required;

WebFreak001: sure, I just tried to keep all the call-sites relatively unchanged since that made it easier to transition to the new function and debug issues, but now that it's working well I can make it better to read and maintain for the future

src/app/organisms/room/msgContent.ts
WebFreak001: I haven't actually tried this out with any videos, but felt that it would make sense to use the same scaling as for images for consistency, since videos are really just moving images. I added a bit larger minimum size in the hope that video controls, at least the play/pause button, are available. Although this isn't tested and may differ from browser to browser anyway.

WebFreak001:
> By collapsing image width we won't be able to distinguish between what is sticker and what is image.

I think this is a good thing, not all matrix clients implement emotes the same way and users might just intend to do this for single-use emotes.

> Making sticker width between 48 to 152 means adding inconsistence UI for sticker with different width, they should follow same size just like custom emoji's.

I think there is value in small stickers. Here is an example from a telegram sticker that I quite enjoy sending, simply because it is unique in being one of few existing small stickers:

image

> And if image/sticker width is really small it clutter up the View image button and image info when Auto Load Media feature is disabled.

ok I tried playing around with this and decided it's probably better to do this in a separate PR since my suggestion would be making the view button smaller / icon-only and moving the info badges to the right of the image box, if it is very narrow.

Keeping the minimum size small, like around 32x32 like it is now, allows people to send small images that look like emotes without making them proper emotes (allowing like one-time emotes) or just having screenshots of small things in an unscaled/undistorted view.

WebFreak001:
narrow image load button proposal 1:

with blur hash:

image

without blur hash:

image

WebFreak001:
PR for narrow images/videos: #1614 

I think we could make it better by making a base class for image/video content box that shares the things such as the badges and file size. Would like to do that in the separate PR if desired though.

ajbura:
Closing as not plained

PR #1554, opened 2023-11-21
Make small images not scale up in image viewer

WebFreak001:
### Description

Makes small images not scale up in image viewer, (when you click an image) instead show them in real resolution, especially good improvement for extra small images and pixel art.

before: image was always scaled up

after: image now in original resolution

(note one is current stable release, other is master, so other things might be slightly different in the screenshots)

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas - n/a
- [ ] I have made corresponding changes to the documentation - n/a
- [x] My changes generate no new warnings

PR #1560, opened 2023-11-26
add large link embeds

WebFreak001:
this PR depends on #1555 and also gets visual improvements from #1554 and #1556, which are already rebased here - but if you don't want to merge those into master I can remove the commits from this PR again as well.

Set as draft while the other PRs are still open, since this PR has the commits from the other PRs included as well (only last commit is relevant to this PR), but some people I chat with wanted to have this early so I already publish it here

This makes images that are sent as links embed as if they were uploaded as images, using the URL preview function. Good for when people are using their own image uploaders or linking to content on the web. Although currently the implementation does not fully mirror how regularly uploaded images look like. Small images are centered inside a static 400px width gray embed box, transparent images also show the gray embed box background.

Clicking on the images to view them fullscreen using the image viewer works well though.

Sample: big previews


cyrneko:
image
Currently with this PR, images will sometimes appear like this.

WebFreak001:
the dependency PRs need to be merged first before this PR can be used or edited further

Can you check out my [`wfr.moe` branch](https://github.com/WebFreak001/cinny/tree/wfr.moe), which has all my local fixes applied, to check if it happens there? (there is also a deployed version on https://cinny.wfr.moe)

cyrneko:
> the dependency PRs need to be merged first before this PR can be used or edited further
> 

Ah, makes sense, I didn't take into account that the preview might not have those.

Works just fine on said branch :)
image

ajbura:
Closing as not plained

PR #1588, opened 2023-12-14
Load assets from relative path

aceArt-GmbH:
### Description

Deploying cinny on a sub path like `/cinny/` breaks loading of some assets.
Use relative path to prevent 404s.

- rendering PDFs
- web app manifest
- web app icons
- web app start

#### Type of change

### Checklist:

### DIFF
--- a/src/app/plugins/pdfjs-dist.ts
+++ b/src/app/plugins/pdfjs-dist.ts
@@ -7,7 +7,7 @@ export const usePdfJSLoader = () =>
   useAsyncCallback(
     useCallback(async () => {
       const pdf = await import('pdfjs-dist');
-      pdf.GlobalWorkerOptions.workerSrc = '/pdf.worker.min.js';
+      pdf.GlobalWorkerOptions.workerSrc = 'pdf.worker.min.js';
       return pdf;
     }, [])
   );
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -6,52 +6,52 @@
   "lang": "en-US",
   "display": "standalone",
   "orientation": "portrait",
-  "start_url": "/",
+  "start_url": "./",
   "background_color": "#fff",
   "theme_color": "#fff",
   "icons": [
     {
-      "src": "/public/android/android-chrome-36x36.png",
+      "src": "./public/android/android-chrome-36x36.png",
       "sizes": "36x36",
       "type": "image/png"
     },
     {
-      "src": "/public/android/android-chrome-48x48.png",
+      "src": "./public/android/android-chrome-48x48.png",
       "sizes": "48x48",
       "type": "image/png"
     },
     {
-      "src": "/public/android/android-chrome-72x72.png",
+      "src": "./public/android/android-chrome-72x72.png",
       "sizes": "72x72",
       "type": "image/png"
     },
     {
-      "src": "/public/android/android-chrome-96x96.png",
+      "src": "./public/android/android-chrome-96x96.png",
       "sizes": "96x96",
       "type": "image/png"
     },
     {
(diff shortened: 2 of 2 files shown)


src/app/plugins/pdfjs-dist.ts
```diff
@@ -7,7 +7,7 @@ export const usePdfJSLoader = () =>
   useAsyncCallback(
     useCallback(async () => {
       const pdf = await import('pdfjs-dist');
-      pdf.GlobalWorkerOptions.workerSrc = '/pdf.worker.min.js';
+      pdf.GlobalWorkerOptions.workerSrc = 'pdf.worker.min.js';
```
ajbura, line 10: As scripts are served from `/assets` i wonder where will it pick pdf.worker.min.js (its in `rootPath/`). `rootPath/pdf.worker.min.js` or `rootPath/assets/pdf.worker.min.js`?

ajbura: for visibility: https://github.com/cinnyapp/cinny/pull/1603/commits/845eebe98af4405bb0840b74d538b5c64522a676 adds build config for providing base url so we can also do like https://github.com/cinnyapp/cinny/pull/1603/commits/845eebe98af4405bb0840b74d538b5c64522a676#diff-fd463d18da0dee07090abb23a56e8e05cf2113a48d1820cb013a5b63d3abd2f9

aceArt-GmbH: After #1603 I need to change `build.config.ts` to include my sub path for index.js to load at all.
Then `pdf.worker.min.js` is loaded absolute from the url. Not absolute to relative path.

Using `${trimTrailingSlash(import.meta.env.BASE_URL)}/pdf.worker.min.js` fixes the problem too.

You decide. I think removing the slash is simpler than trimming and reading env.
`pdf.worker.min.js` is never loaded from `assets/`, it is not even available there

Having to rebuild the app to deploy on a sub-path is not great :/

ajbura: I think removing the slash is fine.

We have to rebuild it because we now have react router and for it have to make all path absolute.

index.html
```diff
@@ -27,7 +27,7 @@
 
     <link id="favicon" rel="shortcut icon" href="./public/favicon.ico" />
 
-    <link rel="manifest" href="/manifest.json" />
+    <link rel="manifest" href="./manifest.json" />
```
ajbura: relative path for manifest is not working if you open https://1588--pr-cinny.netlify.app/login/matrix.org as it search for it at https://1588--pr-cinny.netlify.app/login/manifest.json

absolute path should be prefixed automatically by vite BASE_URL on build, but seems like it's not happening.

ajbura (changes requested):

index.html
aceArt-GmbH: >absolute path should be prefixed automatically by vite BASE_URL on build, but seems like it's not happening.

Yeah, that is probably needed now that we have `/login`.
The `pdf.worker.min.js` will probably need BASE_URL or `import.meta.env.BASE_URL` now too I guess

aceArt-GmbH:
So it seems now the manifest is correctly prefixed if `build.config.ts` is changed.
And `pdf.worker.min.js` is also loaded correctly.
Only the contents inside the manifest are not changed by vite

aceArt-GmbH:
So I just tested again. The paths in the manifest.json don't need further changing (from vite) as they are relative to the manifest itself.
So this should be fine now

ajbura (approved):
Thanks for the changes. LGTM now

Issue #1426, opened 2023-09-13
Further customization of register link


thepaperpilot:
### Describe the problem

I have a cinny instance where I'm disallowing custom home servers and the only home server option uses SSO. The login dialog shows a register link that is effectively non-functional, as the only allowed home server will throw an error when it is clicked.

### Describe the solution you'd like

In this case I'd appreciate being able to make the register link go to a custom page where they can actually register (zitadel in my case). Hiding the text entirely would also work, although suboptimal. 

### Additional context

This problem has been discussed in [this FR](https://github.com/cinnyapp/cinny/issues/146) however the response was that the original FR was implemented, so I'm making this one and I believe it is not a duplicate, or else that PR would have been re-opened.

Issue #1384, opened 2023-08-12
Cannot login to beeper.com homeserver


MFAshby:
### Describe the bug

Cannot login to beeper.com homeserver with cinny

![image](https://github.com/cinnyapp/cinny/assets/1760515/f6735c0b-eebc-4d75-aa19-0b375db61205)

### Reproduction

* go to app.cinny.in
* set homeserver to `beeper.com`

### Expected behavior

expected: login is available

### Platform and versions

```shell
1. OS: linux
2. Browser: firefox
3. Cinny version: v2.2.6
4. Matrix homeserver: beeper.com
```

### Additional context

* I noticed in browser tools that a POST request made to https://matrix.beeper.com/_matrix/client/r0/register fails with 501 (not implemented) error code. This makes sense since registration on beeper.com homeserver is not enabled. I don't know if this complies with the matrix [spec](https://spec.matrix.org/v1.7/client-server-api/#post_matrixclientv3register) so I'll also ask beeper support.
* same homeserver works OK in app.element.io

ajbura:
As a workaround we can password login by typing entire mxid in username field

Issue #1369, opened 2023-08-01
Crash when trying to login to homeserver that does not return `identity_providers` field of `m.login.sso` flow schema


hughns:
### Describe the bug

Cinny is unable to sign in to a Matrix homeserver that offers the `m.login.sso` login type but where the homeserver does not provide a list of identity providers.

Instead the user ends up stuck on a blank screen.

### Reproduction

1. Go to https://app.cinny.in/
2. Enter homeserver `synapse-oidc.lab.element.dev` into the input field
4. You end up on a blank screen
5. Check the Javascript console and see the following error:

```
index-cb0350e9.js:7 TypeError: Cannot read properties of undefined (reading 'sort')
    at hN (index-cb0350e9.js:57022:31396)
    at wy (index-cb0350e9.js:5:52909)
    at Cq (index-cb0350e9.js:9:8732)
    at Aq (index-cb0350e9.js:9:948)
    at Kv (index-cb0350e9.js:9:876)
    at mn (index-cb0350e9.js:9:729)
    at Pk (index-cb0350e9.js:7:10826)
    at index-cb0350e9.js:5:38721
    at g.unstable_runWithPriority (index-cb0350e9.js:1:13085)
    at XE (index-cb0350e9.js:5:38499)
```

### Expected behavior

In the case that the homeserver does not specify a list of identity providers it should show a single SSO login button labelled something similar to "Continue with SSO".

### Platform and versions

```shell
1. OS: macOS
2. Browser: Chrome 115.0.5790.114
3. App: 2.2.6 (app.cinny.in)
4. Homeserver: `synapse-oidc.lab.element.dev`
```

### Additional context

The response to `GET https://synapse-oidc.lab.element.dev/_matrix/client/v3/login` is:

```json
{
    "flows": [
        {
            "type": "m.login.password"
        },
        {
            "type": "m.login.sso",
            "org.matrix.msc3824.delegated_oidc_compatibility": true
        },
        {
            "type": "m.login.token"
        }
    ]
}
```

Which is valid as per the `m.login.sso flow schema` section of https://spec.matrix.org/v1.7/client-server-api/#client-login-via-sso because the `identity_providers` field is optional:

> Optional identity providers (IdPs) to present to the user. These would appear (typically) as distinct buttons for the user to interact with, and would map to the appropriate IdP-dependent redirect endpoint for that IdP.

Issue #1285, opened 2023-06-16
Remove arbitrary restrictions on passwords


omentic:
### Describe the bug

I went to register for a new account on a Matrix server, only to be greeted by the following screen:

![2023-06-15-190237](https://github.com/cinnyapp/cinny/assets/35242550/0624a5d7-ac35-4dd5-9e90-91081e587c93)

This is generally considered bad practice and is not required by the Matrix spec. I suggest removing all restrictions and instead warning for bad practices, similar to what Element does. Though Element also uses a better heuristic: which boils down to length good, dictionary words bad

### Platform and versions

```shell
n/a
```

kfiven:
https://spec.matrix.org/v1.7/client-server-api/#notes-on-password-management says:

> Clients SHOULD enforce that the password provided is suitably complex. The password SHOULD include a lower-case letter, an upper-case letter, a number and a symbol and be at a minimum 8 characters in length. Servers MAY reject weak passwords with an error code M_WEAK_PASSWORD.

omentic:
Oh, that's interesting. That is bad practice and should be removed from the spec, see https://github.com/matrix-org/matrix-spec-proposals/pull/2000#issuecomment-631056746

I wonder why Element doesn't follow that. Maybe someone changed it and forgot the spec mentioned it?

kfiven:
I could find any discussion changing it. Don't know why element is doing that way but correct way to get this fixed everywhere would be by amending the spec.

omentic:
After some discussion in the matrix-spec channel, the consensus is that it's not worth changing it in the spec due to https://github.com/matrix-org/matrix-spec-proposals/pull/3861 coming rather soon. I do think that Cinny should absolutely remove the hard restrictions though - and only *suggest* creating a stronger password via https://github.com/dropbox/zxcvbn (what Element uses) or similar.

Notably the conclusion was that the spec is in fact bad here, for the reasons I linked to in my comment above! The basic idea is that if you are an attacker trying to bruteforce a password, the *only* things that are relevant are 1) length and 2) not being a "common" password, where common means made out of dictionary words or on a common-password list (because that cuts *way* down on the bruteforcing needed).

And from a usability side it's not great because I need to switch to Element just to make my account :slightly_frowning_face:

kfiven:
[...]

PR #1603, opened 2023-12-20
Auth - URL Navigation

ajbura:
### Description

This is first part of making Cinny's layout URL responsive. This PR refactor auth logic and add password reset screen and register with token feature.

#26
#167 
#234 
#1038 
#1269 
#1285 
#1369 
#1384 
#1426

BRAKING CHANGE: self hosted instances now have to configure url re-writes because we are using in-app react-router. example: /netlify.toml

#### Type of change

### Checklist:


config.json
```diff
@@ -1,4 +1,6 @@
 {
+  "appVersion": "3.2.0",
```
WebFreak001: Can you rather make this an optional override instead of putting it into the git config? Otherwise user deployments will need to resolve merge conflicts every time they update cinny - or plainly have a wrong app version in there if they just `cp old-cinny/config.json new-cinny/config.json` from the old deployment.

PR #1617, opened 2024-01-05
remove 0.05em padding from emoji

WebFreak001:
seems completely random and breaks chaining multiple custom emoji together in a row, as well as not being aligned to pixels.

before:

emoji chain

after:

emoji chain

### DIFF
--- a/src/app/styles/CustomHtml.css.ts
+++ b/src/app/styles/CustomHtml.css.ts
@@ -171,7 +171,6 @@ export const EmoticonBase = style([
   DefaultReset,
   {
     display: 'inline-block',
-    padding: '0.05rem',
     height: '1em',
     verticalAlign: 'middle',
   },


ajbura:
The padding between emoji's are on purpose to create good distinction between emoji's. for example two smile should not have their cheeks touching each other or a car and bus emoji put together doesn't look like they are in collision.

WebFreak001:
ok how about me making this a setting then? I personally strongly prefer having emoji touching since you can make creative emoji combinations (and in bridged discord rooms this is also how the other side sees them)

ajbura:
Please create an issue to discuss it.

WebFreak001:
#1634

ajbura:
closing as per https://github.com/cinnyapp/cinny/issues/1634#issuecomment-2582541574

PR #1645, opened 2024-02-25
Add (ctrl-)alt-up/down keyboard navigation

WebFreak001:
### Description

Using Alt-Up/Down moves up/down in the room list / DM list (the drawer)

Using Alt-Left moves up one breadcrumb in a breadcrumbed subspace.

Using Ctrl-Alt-Up/Down moves up/down the pinned spaces (the sidebar) or goes into home or DMs when at the top.

These shortcuts match some other popular messengers (e.g. partially telegram, fully discord, partially element)

Subspaces aren't working optimally yet, since they are immediately entered. In a follow-up PR I would like to improve on that so that they are only selected and not automatically visited when using keyboard navigation, however that requires some more changes I haven't looked into yet.

For navigation the lists or buttons implement imperative refs now. Lists implement a `hasSelected` for the parent to check on which element to start searching for the next item to focus, as well as a `navigate` method which takes in `0`, `1` or `-1`, meaning select current again, select next or select previous respectively. If navigate returns `0`, an item has been selected, if it returns non-zero it means we should continue searching in the adjacent list.

The drawer (home) and sidebar both have an instance of code that goes through a list of lists - could maybe deduplicate this, but without typescript it's a pain to maintain correctness as a public method so I'd rather not do that right now while these components still use plain JS.

#### Type of change

### Checklist:

	- (ran prettier)

### DIFF
--- a/src/app/organisms/navigation/Directs.jsx
+++ b/src/app/organisms/navigation/Directs.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, forwardRef } from 'react';
 import PropTypes from 'prop-types';
 
 import initMatrix from '../../../client/initMatrix';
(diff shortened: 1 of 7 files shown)


ajbura:
closing as outdated because of lack of timely review

PR #1713, opened 2024-04-28
Do not display duplicate reactions

### LINKED ISSUE #1653: Multiple equal reactions from user are displayed multiple times
### Describe the bug

When multiple `m.reaction` events are sent from a user with the same `key` and `rel_type` from the same user, it causes their reactions to add up rather than just being displayed once. Multiple reactions can be caused by buggy clients, multiple clients reacting at the same time, or lag.
image
image

### Reproduction

1. Send a message
2. Get the event id
3. Use element devtools to send this event multiple times of type `m.reaction`:
```json
"m.relates_to": {
  "event_id": "${event_id}",
  "key": "👍",
  "rel_type": "m.annotation"
}
```
4. See issue

### Expected behavior

Only one reaction of a certain emoji should be displayed for a single user.
> Multiple identical annotations (i.e., with the same event type and annotation key) from the same user (i.e., events with the same sender) should be treated as a single annotation.

https://spec.matrix.org/v1.10/client-server-api/#annotations-client-behaviour

### Platform and versions

```shell
- Alpine Linux edge
- Cinny destkop 2.3.1
- Conduit 0.7.0 alpha ([91764745](https://gitlab.com/famedly/conduit/-/commit/9176474513481a035c63c85b7dc6c0025b950dcf) - 2024-03-23)
```

### Additional context

_No response_

TTTaevas:
### Description

"Duplicate reactions" in this context means **reactions using the same key from the same user**

As there are two places in the code from where reactions can be gotten (`Reactions.tsx` & `ReactionViewer.tsx`), the code to fix the problem also appears twice
This code ignores (disables for a line) the `no-param-reassign` ESLint rule, as I couldn't find a way to return something of type `[string, Set<MatrixEvent>]` without returning the parameter itself

I must admit I don't really know if that's the right way to go about it

Fixes #1653

#### Type of change

### Checklist:

### DIFF
--- a/src/app/organisms/room/message/Reactions.tsx
+++ b/src/app/organisms/room/message/Reactions.tsx
@@ -37,7 +37,19 @@ export const Reactions = as<'div', ReactionsProps>(
     const reactions = useRelations(
       relations,
       useCallback((rel) => [...(rel.getSortedAnnotationsByKey() ?? [])], [])
-    );
+    ).map((reaction) => {
+      // Multiple reactions with same user & key may exist
+      // so only show one of them
+      const events = Array.from(reaction[1]);
+      // eslint-disable-next-line no-param-reassign
+      reaction[1] = new Set(
+        events.filter(
+          (event, index) =>
+            events.findIndex((e) => e.sender?.userId === event.sender?.userId) === index
+        )
+      );
+      return reaction;
+    });
 
     const handleViewReaction: MouseEventHandler<HTMLButtonElement> = (evt) => {
       evt.stopPropagation();
--- a/src/app/organisms/room/reaction-viewer/ReactionViewer.tsx
+++ b/src/app/organisms/room/reaction-viewer/ReactionViewer.tsx
@@ -40,7 +40,19 @@ export const ReactionViewer = as<'div', ReactionViewerProps>(
     const reactions = useRelations(
       relations,
       useCallback((rel) => [...(rel.getSortedAnnotationsByKey() ?? [])], [])
-    );
+    ).map((reaction) => {
+      // Multiple reactions with same user & key may exist
+      // so only show one of them
+      const events = Array.from(reaction[1]);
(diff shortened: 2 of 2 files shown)


TTTaevas:

src/app/organisms/room/message/Reactions.tsx
```diff
@@ -37,7 +37,19 @@ export const Reactions = as<'div', ReactionsProps>(
...
-    );
+    ).map((reaction) => {
+      // Multiple reactions with same user & key may exist
+      // so only show one of them
+      const events = Array.from(reaction[1]);
+      // eslint-disable-next-line no-param-reassign
+      reaction[1] = new Set(
+        events.filter(
+          (event, index) =>
+            events.findIndex((e) => e.sender?.userId === event.sender?.userId) === index
```
Vishal-Merugu, line 48: dont use findIndex inside filter use create a map instead to check the existence already

src/app/organisms/room/reaction-viewer/ReactionViewer.tsx
```diff
@@ -40,7 +40,19 @@ export const ReactionViewer = as<'div', ReactionViewerProps>(
...
-    );
+    ).map((reaction) => {
+      // Multiple reactions with same user & key may exist
+      // so only show one of them
+      const events = Array.from(reaction[1]);
+      // eslint-disable-next-line no-param-reassign
+      reaction[1] = new Set(
+        events.filter(
+          (event, index) =>
+            events.findIndex((e) => e.sender?.userId === event.sender?.userId) === index
```
Vishal-Merugu, line 51: same here

Vishal-Merugu (changes requested):
Resolve the comments

TTTaevas:
Can't be bothered to solve conflicting files for a 6-month-old PR that was ignored, feel free to use parts of my code to make your own PR @Vishal-Merugu

PR #1752, opened 2024-05-29
Load manifest.json from relative path

arenekosreal:
PWA is broken without this if deploy cinny on a sub path like `/cinny/` because browser tries to load manifest.json from `example.com/manifest.json` instead `example.com/cinny/manifest.json` and will results to a 404

https://github.com/cinnyapp/cinny/pull/1588 did most of the jobs but it seems that the contributor forgot to update this.

### Description

Fixes #

#### Type of change

### Checklist:

### DIFF
--- a/index.html
+++ b/index.html
@@ -27,7 +27,7 @@
 
     <link id="favicon" rel="shortcut icon" href="./public/favicon.ico" />
 
-    <link rel="manifest" href="/manifest.json" />
+    <link rel="manifest" href="manifest.json" />
     <meta name="mobile-web-app-capable" content="yes" />
     <meta name="application-name" content="Cinny" />
     <meta name="apple-mobile-web-app-title" content="Cinny" />


arenekosreal:

ajbura (commented):
It look like it doesn't work when loading the https://1752--pr-cinny.netlify.app/login/matrix.org as manifest.json is looked relative to current path.

ideally vite should append base url from `build.config.ts` to make absolute url follow configs `/cinny/manifest.json`. But it seems like vite doesn't do that as it doesn't find `manifest.json` at absolute root of the project.

good fix for it would be to make vite append base url on build

arenekosreal:
If append base url on build, the built webapp will not be portable. Maybe I should set `<link rel="manifest"/>` in page and set its `href` attribute dynamically through JavaScript?

ajbura:
to deploy on any subpath (like `/cinny/`) you have to build the app anyway as it is the requirement of app already (because of the react-router-dom).

arenekosreal:
> the built webapp will not be portable

This means that If cinny is packaged by a linux distribution, packagers cannot prepare this for users, and letting users edit built files by themselves is not a good idea.

ajbura:
what i mean is that currently (in dev branch) cinny can only be packaged and distributed for a specific path (`/` officially) and if consumers has to deploy it on other path (say `/cinny/`) they have to build it from source.

arenekosreal:
OK, I will close this PR now because it is only useful for non-supported deployment and also breaks normal functions. Hope someone may find better ways in the future to let both of us happy.

PR #1790, opened 2024-06-30
Add dev branch docker images to registry

Curious-r:
### Description

This PR will provide images with "dev" tag in dockerhub and ghcr. After this change, the "latest" tag in registry will point to the newest release, and "dev" tag will point to the image with the last commit in dev branch. #

#### Type of change

### Checklist:


Curious-r:
PTAL @kfiven

Curious-r:

.github/workflows/dev-deploy.yml
```diff
@@ -0,0 +1,81 @@
+name: Dev branch deploy
+
+on:
+  push:
+    branches:
+      - dev
```
kfiven, line 6: Hi, thanks for working on this. Can we please not run this on every commit as this would be very resource intensive. I think once a week or something similar would be best for this action.

kfiven (changes requested):

.github/workflows/dev-deploy.yml
Curious-r: Sorry for the late reply. I have make an update, add a job to ensure that the interval between image publishes is greater than a week. 
And there is a question: is `DOCKER_PASSWORD` a Docker Hub token? Although it doesn't matter even if it's ineffective, I'm just using it to avoid the rate limit of API requests.
Thanks!

kfiven: Hi, no worries! Can we setup it like a separate action and do the schedule part as explained in [Action docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) as it seems much easier.

Curious-r: 'docker/metadata-action' and 'docker/build-push-action' use git ref of the event that triggers the workflow to automate a lot of things, changing to schedule, we need to configure it in a more complex way.
Or we can just delete `-H "Authorization: Bearer ${{ secrets.DOCKER_PASSWORD }}"` in curl command, our use case doesn't cause rate limiting.

PR #1435, opened 2023-09-25
Prompt to send command as message

cloudrac3r:
### Description

When a message starts with / it triggers the command handler. If no commands are matched, a warning is displayed.

This change allows you to bypass the warning and send the message anyway, including the slash.

Fixes #1434

Screenshot of the modified warning popup, showing a "send as message" button.

Clicking the blue button sends `/i don't know what this is` to the chat. Clicking cancel closes the popup and keeps the text in the chat input.

#### Type of change

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code follows the style guidelines of this project - I hope so! No new warnings in eslint.
- [x] I have performed a self-review of my own code 
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation - I don't think there's anything to say.
- [x] My changes generate no new warnings

### Is this the right way to do it?

Making `processCommand` async and giving it a meaningful return value might be a little hacky? I'm not sure? Advice around code structure would be appreciated, especially as I'm not a React pro.

PR #1919, opened 2024-08-31
Add prompt before sending invalid commands

cloudrac3r:
### Description

This was previously a feature I added in #1435, but it got lost in a refactor. Here it is again.

When a message starts with `/` but didn't trigger a command, the popup will be shown. The popup allows sending it anyway, or going back to edit.

To make this work, I made the submit function async, which does not seem to have had any negative consequences.

This feature will prevent mistakes like this:

image

#### Type of change

### Checklist:

### DIFF
--- a/src/app/components/editor/utils.ts
+++ b/src/app/components/editor/utils.ts
@@ -261,6 +261,17 @@ export const isEmptyEditor = (editor: Editor): boolean => {
   return false;
 };
 
+export const getAttemptedCommand = (editor: Editor): string | undefined => {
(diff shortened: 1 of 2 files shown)


ajbura:
as we now have wysiwyg editor command, emojis, mentions are only appear when they are picked explicitly or selected with the tab press, everything is suppose to work as appear into the editor. so it is not a good idea to add invalid command dialog because even when the command exist with specific prefixed plaintext it wont run unless it appear as cmd pill

PR #1927, opened 2024-09-05
Fix: Prevent marking messages as read when tab inactive

really-not-lavacat:
### Description
When chat in Cinny is opened in browser, but tab isn't active, Cinny marks messages in chat as read anyway, which is really confusing for the conversation partner (for conversation partner it looks like you ignore them).

This PR adds check of `document.hidden`. When tab is inactive, `markAsRead` will do nothing when calling it.

#### Type of change

### Checklist:


really-not-lavacat:

nekoedges (approved):
Looks good, should not make anything bad

greentore:
Doesn't really work for me. I think you should use `document.hasFocus()` instead.

really-not-lavacat:
@greentore done! now `markAsRead` checks both `document.hidden` and `document.hasFocus()` (they're doing different things, so I included both checks)

greentore:
I can't think of situation where `document.hidden` and `document.hasFocus()`  both return `true` at the same time, so the first one is probably redundant.

really-not-lavacat:
When I said `and`, I meant that I used `or` statement, so there's no need for both checks to return `true` at the same time. `document.hidden` and `document.hasFocus()`performing different checks, so I included both.

src/client/action/notifications.ts
```diff
@@ -1,6 +1,10 @@
 import { MatrixClient } from "matrix-js-sdk";
 
 export async function markAsRead(mx: MatrixClient, roomId: string) {
+  // Won't mark messages (or whatever in room) as read when tab inactive
+  // or when browser is inactive (i.e. user interacts with some another app)
+  if (document.hidden || !document.hasFocus()) return;
+
```
ajbura, line 7: It seems like a wrong place to add such conditioning as this function intent is to mark room as read. A good place for require changes would be the code where `markAsRead` is called (i guess in `useEventArrive` hook in `RoomTimeline.tsx` file)

really-not-lavacat: I guess you are right. Thanks for suggestions, I'll rewrite my code when I'll have enough time!

really-not-lavacat:
I don't think I'm gonna make it sometime. Sorry for your wasted time.

PR #1930, opened 2024-09-06
Add authenticated media support

ShadowRZ:
### Description

Implments authenticated media support, which allows Cinny to be used on servers that have frozen unauth media like matrix.org

Fixes #1818

#### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [ ] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings

PR #1946, opened 2024-09-10
Fix video and audio loading with authenicated media

### LINKED ISSUE #1818: Support for MSC3916: Authenticated Media
### Describe the problem

Matrix.org is [planning to sunset the old media endpoints](https://matrix.org/blog/2024/06/26/sunsetting-unauthenticated-media/) in August, which will break media access for clients that do not support [MSC3916](https://github.com/matrix-org/matrix-spec-proposals/pull/3916)

Ensure that Cinny supports Authenticated Media Access according to the MSC

### Alternatives considered

-

### Additional context

-

ShadowRZ:
### Description

Continuation of #1930, Appeareantly Firefox (and maybe Chrome) won't let service workers take over requests from <video> and <audio> tags, so we just fetch the URL ourselves.

Fixes https://github.com/cinnyapp/cinny/issues/1818#issuecomment-2340050747

#### Type of change

### Checklist:

### DIFF
--- a/src/app/components/message/content/util.ts
+++ b/src/app/components/message/content/util.ts
@@ -4,7 +4,8 @@ import { decryptFile } from '../../../utils/matrix';
 export const getFileSrcUrl = async (
   httpUrl: string,
   mimeType: string,
-  encInfo?: EncryptedAttachmentInfo
+  encInfo?: EncryptedAttachmentInfo,
+  forceFetch?: boolean
 ): Promise<string> => {
   if (encInfo) {
     if (typeof httpUrl !== 'string') throw new Error('Malformed event');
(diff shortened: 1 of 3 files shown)


src/app/components/message/content/util.ts
```diff
@@ -4,7 +4,8 @@ import { decryptFile } from '../../../utils/matrix';
 export const getFileSrcUrl = async (
   httpUrl: string,
   mimeType: string,
-  encInfo?: EncryptedAttachmentInfo
+  encInfo?: EncryptedAttachmentInfo,
+  forceFetch?: boolean
```
ajbura, line 8: This function was made to decrypt file but maybe named wrong so I think we should use different function for fetching unencrypted file as adding a boolean param make it hard to reason what this function do.

Apart from it maybe in a different PR we can later remove the line: 15 createObjectURL outside it.

ajbura (changes requested):

src/app/components/message/content/util.ts
ajbura: actually I think I can fix it, I am currently working on fixing the download problem and has to changed getFileSrcUrl usage as I found we already have a `decryptFile` function in app/utils/matrix

ajbura (approved):

PR #1992, opened 2024-10-04
Fix notifications not being sent on decryption correction

### LINKED ISSUE #1892: Notifications not sent on edge case
### Describe the bug

Upon someone sending a message and you opening the room, then unfocusing / changing the room again and someone sending a message with the same amount of highlights in the same room directly afterward, no notification about the message is shown 

### Reproduction

You need two accounts (A1, A2)

1. Open Cinny with A1 while not in the test room
2. Send a message with A2 into the test room (you should get a notification in Cinny)
3. Open the room with A1
4. Close the room again (e.g. click on Home)
5. Send another message with A2 into the test room
6. A1 gets no notification in Cinny

### Expected behavior

A1 should get a notification

### Platform and versions

```shell
1. OS: Arch Linux
2. Browser: LibreWolf
3. Cinny version: https://github.com/cinnyapp/cinny/commit/5058136737e177a5cabecc8b53b489eaabc23429 (dev branch at time of writing)
4. Matrix homeserver: A1: matrix.org, A2: trojaner.dev
```

### Additional context

A friend noticed he didn't get notifications consistently so I dug into the code and found a problem in this snippet:
https://github.com/cinnyapp/cinny/blob/e68c56b334107862ab2eec5635c11c40c050f00b/src/app/pages/client/ClientNonUIFeatures.tsx#L201-L211 

Assume, you get a notification with `unreadInfo = { total: 1, room_id: <test room id>, highlight: 0 }`. `unreadCacheRef` is set to that info directly afterward. So far, so good.

You then read the message and another message is sent into the room. Now, an `unreadInfo` with the same parameters is retrieved, thus the check below is true (as now the `cachedUnreadInfo` is set to `unreadCacheRef`, i.e., the previous `unreadInfo`). So the function returns instead of sending a notification. I could reproduce this exact problem by building from source and setting a debugger in LibreWolf which would, in fact, exit this function exactly there.

TrojanerHD:
### Description

Fixes #1892

My analysis in the issue actually seems to be wrong. The root cause for this issue is not the cache but the fact that the matrix-js-sdk sends a timeline event before calling the fixNotificationCountOnDecryption function (see https://github.com/matrix-org/matrix-js-sdk/blob/da044820d7d3255e41e403bfbd035bf7b90ddd6a/src/client.ts#L10322). This leads to the edge case that cinny returns in the sendTimelineEvent function as the counter is 0 before it is corrected to 1.

#### Type of change

### Checklist:

### DIFF
--- a/src/app/pages/client/ClientNonUIFeatures.tsx
+++ b/src/app/pages/client/ClientNonUIFeatures.tsx
@@ -204,7 +204,14 @@ function MessageNotifications() {
       const cachedUnreadInfo = unreadCacheRef.current.get(room.roomId);
       unreadCacheRef.current.set(room.roomId, unreadInfo);
 
-      if (unreadInfo.total === 0) return;
+      if (unreadInfo.total === 0) {
+        const handleUnreadEvent: RoomEventHandlerMap[RoomEvent.UnreadNotifications] = () => {
+          handleTimelineEvent(mEvent, room, toStartOfTimeline, removed, data);
+          room.removeListener(RoomEvent.UnreadNotifications, handleUnreadEvent);
+        };
+        room.on(RoomEvent.UnreadNotifications, handleUnreadEvent);
+        return;
+      }
       if (
         cachedUnreadInfo &&
         unreadEqual(unreadInfoToUnread(cachedUnreadInfo), unreadInfoToUnread(unreadInfo))


TrojanerHD:

src/app/pages/client/ClientNonUIFeatures.tsx
```diff
@@ -204,7 +204,14 @@ function MessageNotifications() {
       const cachedUnreadInfo = unreadCacheRef.current.get(room.roomId);
       unreadCacheRef.current.set(room.roomId, unreadInfo);
 
-      if (unreadInfo.total === 0) return;
+      if (unreadInfo.total === 0) {
+        const handleUnreadEvent: RoomEventHandlerMap[RoomEvent.UnreadNotifications] = () => {
+          handleTimelineEvent(mEvent, room, toStartOfTimeline, removed, data);
+          room.removeListener(RoomEvent.UnreadNotifications, handleUnreadEvent);
+        };
+        room.on(RoomEvent.UnreadNotifications, handleUnreadEvent);
```
ajbura, line 212: since you are removing the listener inside event handler, it feels like to me that following change may lead to memory leaks where the event handler is set but never gets removed because you never hear from it?

TrojanerHD: Yes, that is something I did not consider. I don't really know how to tackle this - my only idea would be to remove the listener after some time. Or I could change it such that for each room there will be at most one listener which would still add memory overhead but constant in the number of rooms the user joined :thinking:

ajbura:
Closing because of unresolved issue. I have added your findings to the original issue https://github.com/cinnyapp/cinny/issues/1892#issuecomment-3238909930 for later.

PR #2016, opened 2024-10-19
Invitations: show server part from inviter

WebFreak001:
### Description

The invitations list now shows "on **server.domain**" for whoever invited you:

image

Some servers are particularly prone to spam users, this helps identifying spam invites early on.

#### Type of change

### Checklist:

- n/a: I have commented my code, particularly in hard-to-understand areas
- n/a: I have made corresponding changes to the documentation

### DIFF
--- a/src/app/pages/client/inbox/Invites.tsx
+++ b/src/app/pages/client/inbox/Invites.tsx
@@ -32,7 +32,12 @@ import {
 } from '../../../utils/room';
 import { nameInitials } from '../../../utils/common';
 import { RoomAvatar } from '../../../components/room-avatar';
-import { addRoomIdToMDirect, getMxIdLocalPart, guessDmRoomUserId } from '../../../utils/matrix';
+import {
+  addRoomIdToMDirect,
+  getMxIdLocalPart,
+  getMxIdServer,
+  guessDmRoomUserId,
+} from '../../../utils/matrix';
 import { Time } from '../../../components/message';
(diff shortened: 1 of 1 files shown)


src/app/pages/client/inbox/Invites.tsx
```diff
@@ -102,6 +108,12 @@ function InviteCard({ room, userId, direct, compact, onNavigate }: InviteCardPro
         <Box grow="Yes">
           <Text size="T200" priority="300" truncate>
             Invited by <b>{senderName}</b>
+            {serverName && (
+              <>
+                {' on '}
+                <b>{serverName}</b>
+              </>
+            )}
```
ajbura, line 116: I think displaying only `serverName` wont be sufficient here as we only display `display_name` and the username part could be anything. maybe we should show the entire userId as "Invited by __name__ (@ name:server.name)"

WebFreak001: sure this depends just on how much we want to show here, so I can change it to

```suggestion
            {serverName && (
              <>
                {' ('}
                <b>{senderId}</b>
                {')'}
              </>
            )}
```

as well.

Could look a bit more cluttered now, depends on the trade of between form and function that you want to take here. Perhaps we should make this part a bit lighter.

Feel free to just apply the suggestion if you prefer this way.

kfiven:
From the changes after spam invite management, we have full user id in invite, so closing this one.

Issue #1632, opened 2024-01-23
Odd Emoji Autofill Ordering


cyrneko:
### Describe the bug

The way Cinny orders emoji/emotes when searching is quite odd and does not match what most users would expect. For instance, searching for "fire" returns "heart_on_fire", "firefighter" and others before "fire".

### Reproduction

1. Enter any Room
2. Start searching for, or typing the name of an emote/emoji
3. Observe ordering being...odd

### Expected behavior

Generally, I expect there to be a more "exact" order when searching for emotes:
1. Exact Match - First priority, exact matches for the search
2. Users' recently used emotes/emoji
3. Rough Match - essentially fuzzy search

### Platform and versions

```shell
1. Pop!_OS 22.04, Linux 6.6.6
2. LibreWolf 121.0-1
3. Cinny v3.2.0 (https://app.cinny.in)
4. tchncs.de / matrix.tchncs.de
```

ajbura:
accidentally push directly to dev branch, fixed by f73dc05e259cd16145304db0c40415df63d70e3c

cyrneko:
much better, thank you!

PR #2057, opened 2024-11-16
Fix emoji ordering

marekvospel:
### Description
When using emojis in cinny I noticed the order of emojis and stickers being the same as the order I uploaded them. Later when inviting my friends to the room, they showed me, that their order of emojis is different. (across homeservers)
image  
image  

This PR aims to fix this by ordering the emotes alphabetically (which feels like the most logical way to order them) both in emoji picker and inside autocomplete.

Related to #1632 (doesn't fix it, but makes the autocomplete behavior more predictable)

#### Type of change

### Checklist:

### DIFF
--- a/src/app/components/editor/autocomplete/EmoticonAutocomplete.tsx
+++ b/src/app/components/editor/autocomplete/EmoticonAutocomplete.tsx
@@ -58,7 +58,9 @@ export function EmoticonAutocomplete({
   const searchList = useMemo(() => {
     const list: Array<EmoticonSearchItem> = [];
     return list.concat(
-      imagePacks.flatMap((pack) => pack.getImagesFor(PackUsage.Emoticon)),
+      imagePacks.flatMap((pack) =>
+        pack.getImagesFor(PackUsage.Emoticon).sort((a, b) => a.shortcode.localeCompare(b.shortcode))
+      ),
       emojis
     );
   }, [imagePacks]);
--- a/src/app/components/editor/autocomplete/EmoticonAutocomplete.tsx
+++ b/src/app/components/editor/autocomplete/EmoticonAutocomplete.tsx
@@ -62,11 +62,11 @@ export function EmoticonAutocomplete({
         imagePacks.flatMap((pack) => pack.getImagesFor(PackUsage.Emoticon)),
         emojis
       )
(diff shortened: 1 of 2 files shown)


marekvospel:

src/app/components/editor/autocomplete/EmoticonAutocomplete.tsx
```diff
@@ -57,10 +57,12 @@ export function EmoticonAutocomplete({
...
-    );
+    return list
+      .concat(
+        imagePacks.flatMap((pack) => pack.getImagesFor(PackUsage.Emoticon)),
+        emojis
+      )
+      .sort((a, b) => a.shortcode.localeCompare(b.shortcode));
   }, [imagePacks]);
 
   const [result, search, resetSearch] = useAsyncSearch(searchList, getEmoticonStr, SEARCH_OPTIONS);
```
ajbura, line 67: From the performance point of view, I think instead of sorting the `searchList` we should sort the `result.item`.

ajbura (commented):
Overall LGTM.

I think it doesn't fix #1632 specifically as it sort the the search result alphabetically rather than how well they match with the query as per request.
For example searching `fire` result in list containing `heart_on_fire , firefighter` which should be sorted based on the index of `fire`, result word-size and does `fire` is on the word boundary.

marekvospel:
> Overall LGTM.
> 
> I think it doesn't fix #1632 specifically as it sort the the search result alphabetically rather than how well they match with the query as per request.
> For example searching `fire` result in list containing `heart_on_fire , firefighter` which should be sorted based on the index of `fire`, result word-size and does `fire` is on the word boundary.
> 

True, I'll update the description to be more precise.

src/app/components/editor/autocomplete/EmoticonAutocomplete.tsx
marekvospel: Makes sense, I'll update this in about 2 days when I get to my computer.

marekvospel: Sorry for the delay, done as requested

ajbura (approved):
LGTM. Thanks!

Issue #1646, opened 2024-02-26
Feat: MSC2530 support for Media Captions


cyrneko:
### Describe the problem

Cinny currently does not support rendering or sending image captions as per MSC2530. This means:
- Images sent with a caption will not have it visible until hovered to reveal it as Alt-Text
- it isn't possible to send media with a caption
  - Cinny will instead upload the media early, then send both the caption and media in quick succession to emulate media captions.

### Describe the solution you'd like

It would be great if support for [MSC2530](https://github.com/matrix-org/matrix-spec-proposals/pull/2530) could be introduced. This would allow for the above issues to be rectified.

### Alternatives considered

The alternative is to constantly look in the title of an image preview, or hover over an image to get Alt-Text (which is the image caption). This is not ideal as most users would not notice the image caption in these places, and cannot even read it fully in the image viewer's header.

PR #2059, opened 2024-11-17
Render captions to m.file, m.image, m.video, and m.audio

nexy7574:
### Description
This PR will introduce rendering [file captions](https://spec.matrix.org/v1.12/client-server-api/#mimage:~:text=The%20original%20filename%20of%20the%20uploaded%20file.).
These were introduced in spec v1.10, and have recently seen wide adoption amongst Element and other clients. It is time that cinny had this too.

*Sending* captions is [out of scope for this PR](https://github.com/cinnyapp/cinny/pull/2059#discussion_r1883258813) and will be introduced later.

Fixes partially #1646

#### Type of change

- [X] New feature (non-breaking change which adds functionality)

### Checklist:

- [X] I have performed a self-review of my own code [*Image showing two uploads - one with no caption, and one with a caption*](https://files.nexy7574.co.uk/u/XQQKOF.jpg)

### DIFF
--- a/src/app/components/RenderMessageContent.tsx
+++ b/src/app/components/RenderMessageContent.tsx
@@ -69,24 +69,25 @@ export function RenderMessageContent({
     );
   };
   const content: IImageContent = getContent();
-  const renderCaption = content.filename && content.filename !== content.body;
-  let renderedCaption: React.ReactNode = null;
-  if(renderCaption) {
-    renderedCaption = (
-      <MNotice
-        edited={edited}
-        content={getContent()}
-        renderBody={(props) => (
-          <RenderBody
-            {...props}
-            highlightRegex={highlightRegex}
-            htmlReactParserOptions={htmlReactParserOptions}
-            linkifyOpts={linkifyOpts}
-          />
-        )}
-        renderUrlsPreview={urlPreview ? renderUrlsPreview : undefined}
-      />
-    )
+  const renderCaption = () => {
+    if(content.filename && content.filename !== content.body) {
+      return (
+        <MText
+          edited={edited}
+          content={getContent()}
+          renderBody={(props) => (
+            <RenderBody
+              {...props}
+              highlightRegex={highlightRegex}
+              htmlReactParserOptions={htmlReactParserOptions}
+              linkifyOpts={linkifyOpts}
+            />
+          )}
+          renderUrlsPreview={urlPreview ? renderUrlsPreview : undefined}
+        />
+      )
+    }
+    return null;
   }
 
   const renderFile = () => (
@@ -122,7 +123,7 @@ export function RenderMessageContent({
         )}
         outlined={outlineAttachment}
       />
-      {renderedCaption}
+      {renderCaption()}
     </>
   );
 
@@ -196,7 +197,7 @@ export function RenderMessageContent({
           )}
           outlined={outlineAttachment}
         />
-        {renderedCaption}
+        {renderCaption()}
       </>
     );
   }
@@ -231,7 +232,7 @@ export function RenderMessageContent({
           )}
           outlined={outlineAttachment}
         />
-        {renderedCaption}
+        {renderCaption()}
       </>
 
     );
@@ -248,7 +249,7 @@ export function RenderMessageContent({
           )}
           outlined={outlineAttachment}
         />
-        {renderedCaption}
+        {renderCaption()}
       </>
 
     );
--- a/src/app/components/RenderMessageContent.tsx
+++ b/src/app/components/RenderMessageContent.tsx
@@ -68,13 +68,13 @@ export function RenderMessageContent({
       </UrlPreviewHolder>
     );
   };
-  const content: IImageContent = getContent();
   const renderCaption = () => {
+    const content: IImageContent = getContent();
     if(content.filename && content.filename !== content.body) {
       return (
         <MText
           edited={edited}
-          content={getContent()}
+          content={content}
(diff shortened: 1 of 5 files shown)


timedoutuk:

nexy7574:
Deployed live at https://cinny.matrix.nexy7574.co.uk

src/app/components/RenderMessageContent.tsx
```diff
@@ -67,38 +68,62 @@ export function RenderMessageContent({
       </UrlPreviewHolder>
     );
   };
+  const content: IImageContent = getContent();
+  const renderCaption = content.filename && content.filename !== content.body;
+  let renderedCaption: React.ReactNode = null;
+  if(renderCaption) {
+    renderedCaption = (
+      <MNotice
```
ajbura: It will be a good idea if we render it as `m.text`

src/app/components/RenderMessageContent.tsx
```diff
@@ -67,38 +68,62 @@ export function RenderMessageContent({
       </UrlPreviewHolder>
     );
   };
+  const content: IImageContent = getContent();
+  const renderCaption = content.filename && content.filename !== content.body;
+  let renderedCaption: React.ReactNode = null;
+  if(renderCaption) {
```
ajbura: maybe we can put it into a function so it doesn't not float around and rendered when called explicitly.

src/app/features/room/RoomInput.tsx
```diff
@@ -227,17 +227,18 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
       const contentsPromises = uploads.map(async (upload) => {
         const fileItem = selectedFiles.find((f) => f.file === upload.file);
         if (!fileItem) throw new Error('Broken upload');
+        const caption = toPlainText(editor.children).trim() || undefined;
```
ajbura: It looks like following changes add same caption to multiple uploaded file. Do we really want that? Can't we just send it as separate in that case?

ajbura: This only add caption as plain text but spec does support customHTML as well.

src/app/features/room/RoomInput.tsx
```diff
@@ -318,7 +319,9 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
           content['m.relates_to'].is_falling_back = false;
         }
       }
-      mx.sendMessage(roomId, content);
+      if (!uploadBoardHandlers.current) {
+          mx.sendMessage(roomId, content);
+      }
```
ajbura: It will be a good idea if we can extract the `content` generation process in separate function so it can be used here and for caption.

src/app/components/RenderMessageContent.tsx
timedoutuk: Yeah, been using m.notice for a while and it does look a bit out of place. Will update.

src/app/features/room/RoomInput.tsx
timedoutuk: Would it not be better to attach the caption to the last uploaded item, rather than as an entirely separate message?

kfiven:
This should be either a toggle or just rendering support because the way we currently use the Cinny's feature of file upload and sending message togather will not be usefull after this. For example someone wants to send an image right after writing their TWIM message.

Another problem I found is that these captions can not be edited.

src/app/features/room/RoomInput.tsx
kfiven: image
I think captions should be added here for each file and message input should not be used for caption as doing so will break the current functionality. For example someone wants to send an image right after writing their TWIM message.

timedoutuk: As nice as that would be, would it not require adding another message composer, especially since both html and plaintext are meant to be supported? How would that work on mobile?
Apologies for the delay, my hands are quite full until Febuary 😔

ajbura: Yeah adding full message composer won't be a good idea, but maybe we can add plaintext caption as file rename option, not sure about how that would go. So if we don't want to go with sending caption from main composer yet could we please limit this PR for display only support as that alone would be really helpful.

src/app/features/room/RoomInput.tsx
timedoutuk: This code was reverted.

src/app/features/room/RoomInput.tsx
timedoutuk: I have now removed sending captions altogether. Rendering plaintext & HTML captions remains.

For the sake of satisfying other clients, `body` is still the filename (since it's required) but the `filename` field is still sent with the same value.

src/app/components/RenderMessageContent.tsx
```diff
@@ -67,38 +68,63 @@ export function RenderMessageContent({
...
-          renderAsPdfFile={() => (
-            <ReadPdfFile
-              body={body}
-              mimeType={mimeType}
-              url={url}
-              encInfo={encInfo}
-              renderViewer={(p) => <PdfViewer {...p} />}
+  const content: IImageContent = getContent();
+  const renderCaption = () => {
+    if(content.filename && content.filename !== content.body) {
```
ajbura: Looks like `const content: IImageContent...` left to be pulled inside `renderCaption` function?

ajbura (approved):
LGTM. Thanks for the changes!

PR #2071, opened 2024-11-30
Load the local hostname's config, where possible

nexy7574:
### Description
This change mirrors some nice but niche behaviour from Element, wherein `config.$HOSTNAME.json` is tried before `config.json`.
This is very handy, at least for me, because I serve cinny from a multitude of domains, but all from the same dist directory. This means if you visit cinny from one of my domains that is for a different homeserver, you will get the config for only one of my homeservers.
An example of this is https://cinny.transgender.ing - the `config.json` is loaded, which is intended for https://cinny.nexy7574.co.uk. Element, however, nicely tries to load `config.element.transgender.ing.json` in the first case, meaning I can serve a different default login server.

#### Type of change

- [X] New feature (non-breaking change which adds functionality)

### Checklist:

### DIFF
--- a/src/app/components/ClientConfigLoader.tsx
+++ b/src/app/components/ClientConfigLoader.tsx
@@ -4,8 +4,12 @@ import { ClientConfig } from '../hooks/useClientConfig';
 import { trimTrailingSlash } from '../utils/common';
 
 const getClientConfig = async (): Promise<ClientConfig> => {
-  const url = `${trimTrailingSlash(import.meta.env.BASE_URL)}/config.json`;
-  const config = await fetch(url, { method: 'GET' });
+  let url = `${trimTrailingSlash(import.meta.env.BASE_URL)}/config.${window.location.hostname}.json`;
+  let config = await fetch(url, { method: 'GET' });
+  if(config.status===404) {
+    url = `${trimTrailingSlash(import.meta.env.BASE_URL)}/config.json`;
+    config = await fetch(url, { method: 'GET' });
+  }
   return config.json();
 };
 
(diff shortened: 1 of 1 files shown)


src/app/components/ClientConfigLoader.tsx
```diff
@@ -4,8 +4,12 @@ import { ClientConfig } from '../hooks/useClientConfig';
 import { trimTrailingSlash } from '../utils/common';
 
 const getClientConfig = async (): Promise<ClientConfig> => {
-  const url = `${trimTrailingSlash(import.meta.env.BASE_URL)}/config.json`;
-  const config = await fetch(url, { method: 'GET' });
+  let url = `${trimTrailingSlash(import.meta.env.BASE_URL)}/config.${window.location.hostname}.json`;
+  let config = await fetch(url, { method: 'GET' });
+  if(config.status===404) {
```
timedoutuk: It's worth noting that in the dev environment, `npm start` stupidly returns a `200` for this non-existent file. This isn't something that would happen in a production environment (I'd certainly hope at least!), but it does cause the dev server to throw an error while loading.

Might it be worth reading the JSON before returning it (or trying to) to see if the returned config is ACTUALLY 200 OK?

timedoutuk: it doesn't look like Element does this, so if I'm sticking to the original goal of basically just mirroring their behaviour, then this is sufficient.

src/app/components/ClientConfigLoader.tsx
```diff
@@ -4,9 +4,26 @@ import { ClientConfig } from '../hooks/useClientConfig';
...
+
+  if(!loadedConfig) {
+    url = `${trimTrailingSlash(import.meta.env.BASE_URL)}/config.json`;
+    config = await fetch(url, { method: 'GET' });
+    try {
+      loadedConfig = await config.json();
+    } catch (e) {
+      return {};
+    }
+  }
```
ajbura: Getting config blocks the app from loading, so fetching multiple config in series increase the time to interactivity significantly. Please load the standard `base_url/config.json` first or fetch both of them in parallel.

ajbura: Please do not catch (and suppress) the error here as UI is responsible for caching and displaying them to user.

ajbura (changes requested):

ajbura:
Closing it. I feel like it is really complicating the error handling when loading config.

PR #2090, opened 2024-12-14
Fix state of the icons in Toolbar being too hard to see

verbey:
### Description
This makes Cinny use filled variants of icons present in the Toolbar. I believe this is important because as of right now they are only highlighted slightly by the background, which does not differ much ([see this color comparison](https://webaim.org/resources/contrastchecker/?fcolor=404040&bcolor=333333)) from the background of Toolbar itself and thus makes this an accessibility issue for users with low sight like me.
See the relevant PR for the folds repo:
https://github.com/cinnyapp/folds/pull/60
See how this would look like in Cinny with the filled icons:
https://files.catbox.moe/jf3xep.mp4

Fixes icons in Toolbar not having enough visual hints to highlight their state.

#### Type of change

### Checklist:

- [?] I have made corresponding changes to the documentation // Am I supposed to do that for this type of change?

### DIFF
--- a/src/app/components/editor/Editor.css.ts
+++ b/src/app/components/editor/Editor.css.ts
@@ -70,3 +70,11 @@ export const EditorToolbar = style({
 export const MarkdownBtnBox = style({
   paddingRight: config.space.S100,
 });
+
+export const invertOnPress = style({
+  selectors: {
+    '&[aria-pressed="true"]': {
+      filter: 'invert(1)',
+    },
+  },
+});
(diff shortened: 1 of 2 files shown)


ajbura:
preview of change: 
image

ajbura:
Hey, thanks for the changes and concern. In color comparison you are actually comparing the two background colors as `SurfaceVariant` `Active` color is a background color changing on active state of button and foreground color is `OnContainer` which is the color of icon and passes accessibility. 

To improve the visual priority of a button upon pressing, you need to change its variant to something like `Primary`, `Secondary` or `Success` depending on intension behind priority. Inverting the colors of an `IconButton` changes it's design principle(`Iconbutton` can only have `Container` colors as `Main` colors are for call-to-action visuals)

verbey:
Look, a massive blunder on my side happened. The two commits that were pushed into the PR were actually something I was experimenting with but eventually dropped in favor of adding filled versions of icons. Before creating a PR I did a git pull and managed to somehow revert the actual commit with those previous two that were no good in the course of doing a git merge. I am going to close this one and create a proper PR right away. I am sorry for the mistake, it's very late right now and I didn't have enough sleep.

PR #1647, opened 2024-03-01
Add mobile gestures

North-West-Wind:
### Description

I want to use Cinny on my mobile device, but it's pretty "immobile". In order to improve the mobile experience, I added 3 mobile gestures, including:
- Swipe right menu
- Swipe left reply
- Image viewer pan

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] ~~I have made corresponding changes to the documentation~~
- [x] My changes generate no new warnings

PR #2171, opened 2025-01-19
feat: Add mobile gesture (again)

North-West-Wind:
### Description

As Tauri 2.0 released with mobile support, better mobile features seems to be getting more important.  
This is a re-entry of #1647, which was written for Cinny v3, and then Cinny updated to v4 with major refactors and the code no longer works. It's until now I finally have time to rewrite the features, and I hope the same doesn't happen again.

This PR brings in 2 mobile gestures:
- Swipe right menu
- Swipe left reply

#### Type of change

### Checklist:

- [x] ~~I have made corresponding changes to the documentation~~

### DIFF
--- a/src/app/components/BackRouteHandler.tsx
+++ b/src/app/components/BackRouteHandler.tsx
@@ -1,86 +1,10 @@
-import { ReactNode, useCallback } from 'react';
-import { matchPath, useLocation, useNavigate } from 'react-router-dom';
-import {
-  getDirectPath,
-  getExplorePath,
-  getHomePath,
-  getInboxPath,
-  getSpacePath,
-} from '../pages/pathUtils';
-import { DIRECT_PATH, EXPLORE_PATH, HOME_PATH, INBOX_PATH, SPACE_PATH } from '../pages/paths';
+import { ReactNode } from 'react';
+import { useGoBack } from '../hooks/useGoBack';
 
 type BackRouteHandlerProps = {
   children: (onBack: () => void) => ReactNode;
 };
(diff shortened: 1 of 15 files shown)


ajbura:
@North-West-Wind does this fixes any ongoing issue? Cinny now support url navigation which allow user to use mobile native back and forward gesture.

ajbura (commented):
@North-West-Wind does this fixes any ongoing issue? Cinny now support url navigation which allow user to use mobile native back and forward gesture.

North-West-Wind:
I didn't really focus on existing issues.  
I'm aware that Cinny now supports URL navigation, but a slide menu looks nicer.

ajbura:
As it go against the url navigation, I think we should not continue with it as it might lead to further unwanted complication in overall app navigation.

North-West-Wind:
It does use URL navigation though. For example, when using the slide menu in a room, it changes the URL back to the space URL, and the "back" navigation works to bring it back to the room page.

The trigger for URL change is when the slide menu completely opens.

ajbura:
I’ve tested it on mobile, and here are my thoughts:

- The slide animation feels like a drawer(which is fading in mobile world), which causes the back screen to appear after the animation. This conflicts with the native back gestures on mobile and browser.
- The swipe-left-to-reply functionality causes the content to shift left when scrolling back.

Given the trade-off between the value this change provides and the user experience impact, I think it might not be worth implementing.

North-West-Wind:
Hmm, well then. I'll see if I have time to work on it in the future. Thank you for reviewing this :>

ajbura:
closing as not planned

PR #2175, opened 2025-01-21
Fix 'enter' immediately sending the message using IME keyboard on Safari

### LINKED ISSUE #2103: When in Japanese input, 'enter' should not immediately send the message
### Describe the bug

In Japanese IME, you typically type text, then hit enter to "commit" the correct Kanji. For example, if I type テスト then it will be underlined until I hit 'enter', in effect confirming that I've finished writing and picked the correct Kanji/characters

Cinny binds to enter, so even though I am not done typing and only committed a word, as soon as I hit enter, Cinny sends the message 

It's a very common problem when using `onkeyup` as listener, eg: https://stackoverflow.com/questions/46812908/enter-keyup-event-in-japanese-input

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
macOS Desktop 4.2.3, macOS 15.2
```

### Additional context

_No response_

programmablereya:
### Description

On most browsers, pressing Enter to end IME composition produces this sequence of events:
* keydown (keycode 229, key Processing/Unidentified, isComposing true)
* compositionend
* keyup (keycode 13, key Enter, isComposing false)

On Safari, the sequence is different:
* compositionend
* keydown (keycode 229, key Enter, isComposing false)
* keyup (keycode 13, key Enter, isComposing false)

This causes Safari users to mistakenly send their messages when they press Enter to confirm their choice in an IME.

The workaround is to treat the next keydown with keycode 229 as if it were part of the IME composition period if it occurs within a short time of the compositionend event.

Fixes #2103, but needs confirmation from a Safari user.

#### Type of change

### Checklist:

### DIFF
--- a/src/app/state/lastCompositionEnd.ts
+++ b/src/app/state/lastCompositionEnd.ts
@@ -1,0 +1,3 @@
+import { atom } from 'jotai';
+
+export const lastCompositionEndAtom = atom<number | undefined>(undefined);
--- a/src/app/features/room/RoomInput.tsx
+++ b/src/app/features/room/RoomInput.tsx
@@ -116,6 +116,7 @@ import { useRoomCreators } from '../../hooks/useRoomCreators';
 import { useTheme } from '../../hooks/useTheme';
 import { useRoomCreatorsTag } from '../../hooks/useRoomCreatorsTag';
 import { usePowerLevelTags } from '../../hooks/usePowerLevelTags';
+import { useComposingCheck } from '../../hooks/useComposingCheck';
 
 interface RoomInputProps {
   editor: Editor;
@@ -216,6 +217,8 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
     const handlePaste = useFilePasteHandler(handleFiles);
     const dropZoneVisible = useFileDropZone(fileDropContainerRef, handleFiles);
     const [hideStickerBtn, setHideStickerBtn] = useState(document.body.clientWidth < 500);
+
+    const isComposing = useComposingCheck();
 
     useElementSizeObserver(
       useCallback(() => document.body, []),
@@ -380,7 +383,7 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
       (evt) => {
         if (
           (isKeyHotkey('mod+enter', evt) || (!enterForNewline && isKeyHotkey('enter', evt))) &&
-          !evt.nativeEvent.isComposing
+          !isComposing(evt)
         ) {
           evt.preventDefault();
           submit();
@@ -394,7 +397,7 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
           setReplyDraft(undefined);
         }
       },
-      [submit, setReplyDraft, enterForNewline, autocompleteQuery]
+      [submit, setReplyDraft, enterForNewline, autocompleteQuery, isComposing]
     );
 
     const handleKeyUp: KeyboardEventHandler = useCallback(
--- a/src/app/features/room/message/MessageEditor.tsx
+++ b/src/app/features/room/message/MessageEditor.tsx
@@ -53,6 +53,7 @@ import { AsyncStatus, useAsyncCallback } from '../../../hooks/useAsyncCallback';
 import { useMatrixClient } from '../../../hooks/useMatrixClient';
 import { getEditedEvent, getMentionContent, trimReplyFromFormattedBody } from '../../../utils/room';
 import { mobileOrTablet } from '../../../utils/user-agent';
+import { useComposingCheck } from '../../../hooks/useComposingCheck';
 
 type MessageEditorProps = {
   roomId: string;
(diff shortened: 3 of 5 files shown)


programmablereya:

src/app/organisms/settings/Settings.jsx
```diff
@@ -47,6 +47,7 @@ import { settingsAtom } from '../../state/settings';
 import { isMacOS } from '../../utils/user-agent';
 import { KeySymbol } from '../../utils/key-symbol';
 import { useMatrixClient } from '../../hooks/useMatrixClient';
+import { isComposing } from '../../utils/keyboard.js';
```
dvcrn:
```suggestion
import { isComposing } from '../../utils/keyboard;
```

src/app/organisms/search/Search.jsx
```diff
@@ -27,6 +27,7 @@ import { useKeyDown } from '../../hooks/useKeyDown';
 import { openSearch } from '../../../client/action/navigation';
 import { useMatrixClient } from '../../hooks/useMatrixClient';
 import { factoryRoomIdByActivity } from '../../utils/sort';
+import { isComposing } from '../../utils/keyboard.js';
```
dvcrn:
```suggestion
import { isComposing } from '../../utils/keyboard';
```

dvcrn (commented):
Pathes for these 2 imports is incorrect so it can't get resolved properly 

Besides that, this PR fixes IME errors for me. I can type in Japanese normally ❤️

dvcrn:
I've merged this into my fork of Cinny https://github.com/dvcrn/davids-cinny/releases/tag/v4.4.0

src/app/utils/keyboard.ts
```diff
@@ -11,23 +12,36 @@ export interface KeyboardEventLike {
...
-  if (isKeyHotkey('tab', evt)) {
+  if (!isComposing(evt) && isKeyHotkey('tab', evt)) {
     evt.preventDefault();
     callback();
   }
 };
 
 export const preventScrollWithArrowKey: KeyboardEventHandler = (evt) => {
-  if (isKeyHotkey(['arrowup', 'arrowright', 'arrowdown', 'arrowleft'], evt)) {
+  if (!isComposing(evt.nativeEvent) && isKeyHotkey(['arrowup', 'arrowright', 'arrowdown', 'arrowleft'], evt)) {
```
BLumia: Is this change necessary? I've tested it with Chrome + Chinese IME and try to press arrow keys while composing, and it seems this change is not needed. Or maybe I missed something?

src/app/pages/auth/ServerPicker.tsx
```diff
@@ -53,6 +53,9 @@ export function ServerPicker({
   };
 
   const handleKeyDown: KeyboardEventHandler<HTMLInputElement> = (evt) => {
+    if (isComposing(evt.nativeEvent)) {
```
BLumia: Shouldn't this (and also some other `isComposing()` calls) be `isComposing(evt)` instead?

src/app/hooks/useSafariCompositionTaggingForKeyDown.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    function recordCompositionEnd(evt: CompositionEvent) {
+      compositionJustEndedAt = evt.timeStamp
+    }
+
+    function interceptAndTagKeyDown(evt: KeyboardEvent) {
+      if (compositionJustEndedAt !== null
+        && evt.keyCode === 229
+        && (evt.timeStamp - compositionJustEndedAt) < compositionEndThreshold) {
+        Object.assign(evt, { [actuallyComposingTag]: true })
```
ajbura: `actuallyComposingTag` is the variable outside the scope of this hook and is side effect, the `actuallyComposingTag` variable is a good candidate for the /app/state as atom if you want to retain state and available throughout app.

src/app/utils/keyboard.ts
```diff
@@ -1,5 +1,6 @@
 import { isKeyHotkey } from 'is-hotkey';
 import { KeyboardEventHandler } from 'react';
+import { isTaggedAsComposing } from '../hooks/useSafariCompositionTaggingForKeyDown';
```
ajbura: importing from hooks to utils is discouraged to prevent any unknown circular dependencies formation. will be a good idea to pass the data to this func as arguments.

ajbura (changes requested):
Apart from the requested changes is it really needed to check for it everywhere listening to keyboard event as the original issue only request for fix in the message editor specifically?

programmablereya:
Total overhaul to make a much smaller change that uses atoms - how does this look?

src/app/state/lastCompositionEnd.ts
```diff
@@ -0,0 +1,3 @@
+import { atom } from 'jotai';
+
+export const lastCompositionEndAtom = atom<number>(Number.NEGATIVE_INFINITY);
```
ajbura: since we do not know the initial value could we set it to `undefined`

ajbura (changes requested):
Looks good overall. Thanks for the changes!

src/app/state/lastCompositionEnd.ts
programmablereya: It is done!

kfiven (approved):

ajbura (changes requested):
Can you please prettify your change as it seems they are not prettified? It will create unnecessary changes for other contributors who work with prettier.

programmablereya:
> Can you please prettify your change as it seems they are not prettified? It will create unnecessary changes for other contributors who work with prettier.

Done.

ajbura (approved):
LGTM. Thanks!

PR #2212, opened 2025-02-15
Add support for spoilers on images (MSC4193)

### LINKED ISSUE #1966: Implement MSC4193
### Describe the problem

There is not a way to add a spoiler/CW (content warning) to media.

Implement [MSC4193](https://github.com/matrix-org/matrix-spec-proposals/pull/4193) instead.

### Alternatives considered

_No response_

### Additional context

I may be biased on this suggestion, as I made the MSC.

gingershaped:
### Description

Adds support for [MSC4193](https://github.com/matrix-org/matrix-spec-proposals/pull/4193) by allowing users to mark image attachments as spoilers, which will blur them in the UI until explicitly revealed. The `m.spoiler.reason` field is supported for received images, although this PR does not add a way to set it for sent images.

Fixes #1966 

#### Type of change

### Checklist:


gingershaped:
If the most recent message in a room's history is a spoilered image which has been revealed, it will be re-hidden when the next message is sent. I'm not sure why this happens, but it seems to be a symptom of a bug which is out of the scope of this PR to fix.

gingershaped:

src/app/components/message/content/ImageContent.tsx
```diff
@@ -77,6 +81,7 @@ export const ImageContent = as<'div', ImageContentProps>(
     const [load, setLoad] = useState(false);
     const [error, setError] = useState(false);
     const [viewer, setViewer] = useState(false);
+    const [spoiled, setSpoiled] = useState(spoiler ?? false);
```
ajbura: I think the variable has inverse input as per its label. it should be `useState(spoiler ? false : true)`

src/app/components/message/content/ImageContent.tsx
```diff
@@ -160,7 +165,7 @@ export const ImageContent = as<'div', ImageContentProps>(
           </Box>
         )}
         {srcState.status === AsyncStatus.Success && (
-          <Box className={css.AbsoluteContainer}>
+          <Box className={classNames(css.AbsoluteContainer, spoiled && css.Blur)}>
```
ajbura: as per the variable name it should be `spoiled ? undefined : css.Blur`

src/app/components/message/content/ImageContent.tsx
```diff
@@ -172,6 +177,35 @@ export const ImageContent = as<'div', ImageContentProps>(
             })}
           </Box>
         )}
+        {srcState.status === AsyncStatus.Success && spoiled && (
```
ajbura: and here it should be `spoiler && !spoiled`.

src/app/components/message/content/ImageContent.tsx
```diff
@@ -172,6 +177,35 @@ export const ImageContent = as<'div', ImageContentProps>(
...
+              position="Top"
+              align="Center"
+            >
+              {(triggerRef) => (
+                <Button
+                  ref={triggerRef}
+                  variant="Secondary"
+                  fill="Solid"
+                  radii="300"
+                  size="300"
```
ajbura: I think we should instead use a `Chip` of `500` size here (because media autoload disable shows a button) and probably giving it a `Warning` variant with `Pill` radii.

src/app/components/message/content/ImageContent.tsx
```diff
@@ -172,6 +177,35 @@ export const ImageContent = as<'div', ImageContentProps>(
...
+                  onClick={() => setSpoiled(false)}
+                  before={<Icon size="Inherit" src={Icons.Eye} filled />}
```
ajbura: Should not use Eye icon as it is not toggleable button.

src/app/components/message/content/ImageContent.tsx
```diff
@@ -172,6 +177,35 @@ export const ImageContent = as<'div', ImageContentProps>(
...
+                >
+                  <Text size="B300">Show</Text>
```
ajbura: Text for this should be "Spoiler" which better reflect what it is.

src/app/components/upload-card/UploadCardRenderer.tsx
```diff
@@ -53,6 +66,30 @@ export function UploadCardRenderer({
               <Text size="B300">Retry</Text>
             </Chip>
           )}
+          {file.type.startsWith('image') && (
+            <TooltipProvider
+              tooltip={
+                <Tooltip variant="SurfaceVariant">
+                  <Text>Spoil Image</Text>
```
ajbura: It should be "Spoiler Image"

src/app/state/room/roomInputDrafts.ts
```diff
@@ -3,22 +3,66 @@ import { atomFamily } from 'jotai/utils';
 import { Descendant } from 'slate';
 import { EncryptedAttachmentInfo } from 'browser-encrypt-attachment';
 import { IEventRelation } from 'matrix-js-sdk';
-import { TListAtom, createListAtom } from '../list';
 import { createUploadAtomFamily } from '../upload';
 import { TUploadContent } from '../../utils/matrix';
 
-export const roomUploadAtomFamily = createUploadAtomFamily();
+export type TUploadMetadata = {
+  spoiled: boolean;
```
ajbura: It should be named `spoiler` i think.

ajbura (changes requested):
Apart from the request  changes, We should not delete `list.ts` as it is a generic atom file which can be used somewhere else, or even better maybe we can change in "PUT" action to update the existing entry if it already exist.

src/app/components/message/content/ImageContent.tsx
gingershaped: I've kept the variant as-is because the amber color of the `Warning` variant doesn't stand out against similarly-colored spoiled images.

src/app/components/message/content/ImageContent.tsx
```diff
@@ -172,6 +178,34 @@ export const ImageContent = as<'div', ImageContentProps>(
...
+                <Chip
+                  ref={triggerRef}
+                  variant="Secondary"
+                  radii="Pill"
+                  size="500"
+                  outlined
+                  onClick={() => setBlurred(false)}
+                >
+                  <Text size="B300">Spoiler</Text>
+                </Chip>
```
ajbura: I din't think about that the `Chip` doesn't have `fill="Solid"`, your previous choice was actually good.

But I was thinking if we can eliminate the double click when auto load media is disabled. What if we show the spoiler button only when auto load is enabled and on the other case if image has spoiler we can render Spoiler button instead of view button?

Also currently loading spinner overlaps this button when image loads, so maybe instead of blurring the image, what if we do "disable media autoload" thing when image is a spoiler regardless of state of the "disable media autoload" feature. let me know if that make sense to you?

src/app/components/upload-card/UploadCardRenderer.tsx
```diff
@@ -53,6 +66,30 @@ export function UploadCardRenderer({
...
+              {(triggerRef) => (
+                <IconButton
+                  ref={triggerRef}
+                  onClick={toggleSpoiler}
+                  aria-label="Spoil Image"
+                  variant="SurfaceVariant"
+                  radii="Pill"
+                  size="300"
+                >
+                  <Icon src={metadata.markedAsSpoiler ? Icons.EyeBlind : Icons.Eye} size="200" />
```
ajbura: if found that changing the icon makes it a little confusing about the state of the spoiler. we should make it similar to how we have spoiler button in rich editor (Icons.EyeBlind with pressed state when acitve)

ajbura (changes requested):

src/app/components/message/content/ImageContent.tsx
gingershaped: I think having the blur effect makes it more obvious what's going on, so I'll just have the spoiler button override the loading spinner (this is also how Discord handles it.)

src/app/components/upload-card/UploadCardRenderer.tsx
```diff
@@ -53,6 +66,31 @@ export function UploadCardRenderer({
               <Text size="B300">Retry</Text>
             </Chip>
           )}
+          {file.type.startsWith('image') && (
+            <TooltipProvider
+              tooltip={
+                <Tooltip variant="SurfaceVariant">
+                  <Text>Spoiler Image</Text>
```
kfiven:
```suggestion
                  <Text>Mark as Spoiler</Text>
```

src/app/components/upload-card/UploadCardRenderer.tsx
```diff
@@ -53,6 +66,31 @@ export function UploadCardRenderer({
...
+                </Tooltip>
+              }
+              position="Top"
+              align="Center"
+            >
+              {(triggerRef) => (
+                <IconButton
+                  ref={triggerRef}
+                  onClick={toggleSpoiler}
+                  aria-label="Spoil Image"
```
kfiven:
```suggestion
                  aria-label="Mark as Spoiler"
```

src/app/components/message/content/ImageContent.tsx
```diff
@@ -172,8 +178,42 @@ export const ImageContent = as<'div', ImageContentProps>(
             })}
           </Box>
         )}
+        {blurred && (
```
kfiven: image

Spoiler button is also visible if media fails to load.

kfiven (changes requested):

kfiven (approved):
Thanks!

ajbura (approved):

PR #2227, opened 2025-02-24
add example caddyfile

mininmobile:
as i was confused on how to configure regular routing with caddy before, i hope this will help others too ^^

### DIFF
--- a/contrib/caddy/caddyfile
+++ b/contrib/caddy/caddyfile
@@ -1,0 +1,6 @@
+cinny.domain.tld {
+  @nativeRouter not file {path} /
+  rewrite @nativeRouter {http.matchers.file.relative}
+  root * /path/to/caddy/dist
+  file_server
+}


mininmobile:
I have read the CLA Document and I hereby sign the CLA ✅

kfiven:
recheck

contrib/caddy/caddyfile
```diff
@@ -0,0 +1,6 @@
+cinny.domain.tld {
+  @nativeRouter not file {path} /
+  rewrite @nativeRouter {http.matchers.file.relative}
+  root * /path/to/caddy/dist
+  file_server
+}
```
kfiven, line 6: Hi, I am not familiar with caddy, but could you please clarify that,

Does this also set redirects for files like [the ngnix](https://github.com/cinnyapp/cinny/blob/b7e5e0db3e135534b33896f5de139ef8331f1e30/contrib/nginx/cinny.domain.tld.conf#L21-L33) file does?

mininmobile: it's much simpler, it redirects any requested path that does not exist in the directory to `/` (`index.html`)

mininmobile: a more certain approach would be to list out all the files that shouldn't be redirected like in the nginx config (with a `not path /manifest.json /config.json etc...` matcher) but that would require being manually updated were cinny to change what files it needs to work

kfiven: Lets keep it simple then, thanks for explaining.

kfiven (approved):

PR #2266, opened 2025-03-10
fix press enter while composing will cause send unfinished message

BLumia:
### Description

When using an IME (Chinese or Japanese for example), one might use <kbd>Enter</kbd> to commit the text from the IME. Currently Cinny will attempt to send the text from the input box even if the user is still using an IME (i.e. while composing)

#### Type of change

### Checklist:

### DIFF
--- a/src/app/features/room/RoomInput.tsx
+++ b/src/app/features/room/RoomInput.tsx
@@ -360,7 +360,7 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
 
     const handleKeyDown: KeyboardEventHandler = useCallback(
       (evt) => {
-        if (isKeyHotkey('mod+enter', evt) || (!enterForNewline && isKeyHotkey('enter', evt))) {
+        if ((isKeyHotkey('mod+enter', evt) || (!enterForNewline && isKeyHotkey('enter', evt))) && !evt.nativeEvent.isComposing) {
           evt.preventDefault();
           submit();
         }
--- a/src/app/features/room/message/MessageEditor.tsx
+++ b/src/app/features/room/message/MessageEditor.tsx
@@ -163,7 +163,7 @@ export const MessageEditor = as<'div', MessageEditorProps>(
 
     const handleKeyDown: KeyboardEventHandler = useCallback(
       (evt) => {
-        if (isKeyHotkey('mod+enter', evt) || (!enterForNewline && isKeyHotkey('enter', evt))) {
+        if ((isKeyHotkey('mod+enter', evt) || (!enterForNewline && isKeyHotkey('enter', evt))) && !evt.nativeEvent.isComposing) {
           evt.preventDefault();
           handleSave();
         }


BLumia:

kfiven:
Hi @BLumia , How this can be tested? I have Windows/Android devices.

BLumia:
> Hi @BLumia , How this can be tested? I have Windows/Android devices.

Hi!

I guess this might probably not needed to be tested on Android devices. For Windows (and also Linux and macOS), you can install an input method for Chinese (or Japanese), switch to that IME, type something in the input area (then the IME dropdown will be shown), and then press Enter.

If you don't know any IME to install, I suggest the built-in one from Windows: https://support.microsoft.com/en-us/windows/microsoft-simplified-chinese-ime-9b962a3b-2fa4-4f37-811c-b1886320dd72

Alternatively, you can try fcitx-rime: https://rime.im/download/

If needed, I can also attach a video showcase how I tested my fix locally. Let me know if it's needed :)

---

Update:

Here are some videos that can be used as a reference. The video were recorded on a macOS device with [fcitx-macos](https://github.com/fcitx-contrib/fcitx5-macos), but the same steps applies to Windows as well:

Before the fix (<kbd>Enter</kbd> was pressed at 0:05):

https://github.com/user-attachments/assets/673d735d-6f3b-45ae-8f09-35c7e6c73dd7

After the fix (<kbd>Enter</kbd> was pressed at 0:04 to submit the inputted text as-is, and pressed at 0:05 again to send the message):

https://github.com/user-attachments/assets/9495547f-c536-4eee-b3d5-83712f85ce9f

BLumia:
@kfiven Hi, did you test this PR? Do you need any assistant for testing this PR? Let me know if you need any help so hopefully this PR can be merged to benefit users who need to work with IME :)

kfiven:
Hi @BLumia , Unfortunately I didn't got the time to test but just saw your video and there seems to be no issues. One thing I am worried about is how that will interact with the option of disable sending message on Enter in settings?

BLumia:
> One thing I am worried about is how that will interact with the option of disable sending message on Enter in settings?

Well, it's actually not related. For languages that need an IME to input, the activated IME should handle key events beforehand only when the IME is in compose mode. "Sending message on Enter" option should be an option that works when IME is not activated (i.e. in compose mode).

Assume the IME is enabled for a Chinese user, when user type the following keys (note that there are two `⏎`s, and we will focus on the second one):

```plain
example⏎⏎
```

When user input `example` *without* pressing `⏎`, it looks like (screenshot on Windows with Microsoft Pinyin IME):

image

As you can see the IME bar, it means you are in compose mode.

When the user press `⏎` at this exact time, the text will be submitted as-is, and the IME bar will disappear (because you pressed `⏎`), after this you are no longer in compose mode.

Then if you press `⏎` again, this time the application will handle the `⏎` event. In cinny's case, if "Sending message on Enter" is on, the message will be sent, otherwise it will enter a new line. This is also how it works in other IM applications, like Discord and Tencent QQ for example.

kfiven (approved):
LGTM, thanks!

kfiven:
Wondering if it fixes #2103 ?

BLumia:
Oh, just found out the issue you linked. Here is my answer/conclusion after checking that issue:

> Wondering if it fixes https://github.com/cinnyapp/cinny/issues/2103 ?

Well, yes and no.

It should be fixed for users who don't use Safari. According to [this comment](https://github.com/cinnyapp/cinny/issues/2103#issuecomment-2605702031) it seems Safari sends `compositionend` before sending the Enter `keydown` event. All other browsers I've tested (Chrome and Firefox under macOS, Linux and Windows) should work fine.

I don't have access to macOS devices recently and I personally don't use Safari, so I don't plan to do Safari-specific hacks at least for now. For Safari users, the issue will persist. For other Chrome and Firefox-based browser users, this issue should be gone.

I also checked the other PR linked in that issue (https://github.com/cinnyapp/cinny/pull/2175), which seems good, but also *seems* have some unnecessary changes (for example, we probably don't need to add the `isComposing` check for `preventScrollWithArrowKey`, and we already assumed `'nativeEvent' in evt` in some other places so it probably also not needed to be checked inside `isComposing`).

I personally suggest merging this one first since only two lines are changed to fix the issue for most users, then we can review and hopefully merge https://github.com/cinnyapp/cinny/pull/2175 for Safari users as well :)

PR #2268, opened 2025-03-10
Allow more servers to be added to the list in the explore tab

gingershaped:
### Description
This PR saves servers added to the Explore tab in account storage, allowing the list of explorable servers to be customized by the user.

Currently there is no way to explore a server without saving it to the list. Perhaps another button could be added below the current "Add Server" button to view a server without saving it?

#### Type of change

### Checklist:


src/app/hooks/useExploreServers.ts
```diff
@@ -0,0 +1,38 @@
...
+export const useExploreServers = (): [string[], (action: ExploreServerListAction) => void] => {
+  const mx = useMatrixClient();
+  const userAddedServers =
+    useAccountData(AccountDataEvent.CinnyExploreServers)?.getContent<InCinnyExploreServersContent>()
+      ?.servers ?? [];
+
+  const setUserAddedServers = (action: ExploreServerListAction) => {
+    if (action.type === 'APPEND') {
+      mx.setAccountData(AccountDataEvent.CinnyExploreServers, {
+        servers: [...userAddedServers, action.server],
```
ajbura: here we should make sure that a server can not have duplicated entries.

src/app/hooks/useExploreServers.ts
```diff
@@ -0,0 +1,38 @@
...
+      server: string;
+    };
+
+export const useExploreServers = (): [string[], (action: ExploreServerListAction) => void] => {
+  const mx = useMatrixClient();
+  const userAddedServers =
+    useAccountData(AccountDataEvent.CinnyExploreServers)?.getContent<InCinnyExploreServersContent>()
+      ?.servers ?? [];
+
+  const setUserAddedServers = (action: ExploreServerListAction) => {
```
ajbura: instead of using the action we should make two separate functions i.e. `addServer` and `removeServer`.

ajbura: function should be wrapped inside `useCallback` hook so when parent component re-render these function do not cause rerun effect in consumer code.

src/app/pages/client/explore/Explore.tsx
```diff
@@ -131,7 +130,7 @@ export function AddServer() {
                     <Text size="B400">Save</Text>
                   </Button> */}
 
-                  <Button type="submit" onClick={handleView} variant="Secondary" fill="Soft">
+                  <Button type="submit" onClick={handleSubmit} variant="Secondary" fill="Soft">
```
ajbura: Yes, we should also add view button for viewing without saving the server in list.

src/app/hooks/useExploreServers.ts
ajbura: those should also return a promise so button click can show the spinner animation while the request complete.

src/app/pages/client/explore/Explore.tsx
```diff
@@ -155,14 +154,62 @@ export function AddServer() {
...
+              </Text>
+            </Box>
+          </Box>
+        </NavItemContent>
+      </NavLink>
+      {onRemove !== null && hover && (
+        <NavItemOptions>
+          <IconButton onClick={onRemove} variant="Background" fill="None" size="300" radii="300">
+            <Icon size="50" src={Icons.Minus} />
+          </IconButton>
```
ajbura: we should track remove progress and show a spinner here

src/types/matrix/accountData.ts
```diff
@@ -4,6 +4,7 @@ export enum AccountDataEvent {
   IgnoredUserList = 'm.ignored_user_list',
 
   CinnySpaces = 'in.cinny.spaces',
+  CinnyExploreServers = 'in.cinny.explore_servers',
```
ajbura: I think naming it `in.cinny.explore` is good, as the content has `servers` array and allows `content` to add more things outside just servers.

ajbura (changes requested):

.vscode/settings.json
```diff
@@ -1,5 +1,8 @@
 {
   "editor.formatOnSave": true,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "[typescript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  }
```
ajbura, line 7: why the `editor.defaultFormatter` is set in typescript specific rule, when at line 3 it is used for all languages which include typescript. isn't it redundant?

src/app/hooks/router/useExploreSelected.ts
```diff
@@ -11,7 +11,7 @@ export const useExploreSelected = (): boolean => {
   return !!match;
 };
 
-export const useExploreFeaturedSelected = (): boolean => {
+export const useExploringFeaturedRooms = (): boolean => {
```
ajbura: is there any specific reason for changing it's name? previous naming strategy is used because it clearly describe the section Explore > Featured (selected usage).

src/app/pages/client/explore/Server.tsx
```diff
@@ -512,7 +533,77 @@ export function PublicRooms() {
                 {server}
               </Text>
             </Box>
-            <Box grow="Yes" basis="No" />
+            <Box
+              shrink="No"
+              grow={screenSize === ScreenSize.Mobile ? 'No' : 'Yes'}
+              basis={screenSize === ScreenSize.Mobile ? 'Yes' : 'No'}
+              justifyContent="End"
+            >
```
ajbura: it looks like the server name doesn't align properly in center on mobile screen size

src/app/pages/client/explore/Server.tsx
```diff
@@ -512,7 +533,77 @@ export function PublicRooms() {
...
+                        <MenuItem
+                          onClick={handleRemoveServer}
+                          variant="Critical"
+                          fill="None"
+                          size="300"
+                          after={
+                            isRemoving ? (
+                              <Spinner fill="Solid" variant="Secondary" size="200" />
+                            ) : (
+                              <Icon size="100" src={Icons.Minus} />
```
ajbura: it will be more clear using `Delete`(bin) icon instead of `Minus`.

src/app/pages/client/explore/Explore.tsx
```diff
@@ -37,40 +41,41 @@ import { AsyncStatus, useAsyncCallback } from '../../../hooks/useAsyncCallback';
...
-    setDialog(false);
-  };
-
-  const handleView = () => {
-    const server = getInputServer();
-    if (!server) return;
-    navigate(getExploreServerPath(server));
-    setDialog(false);
-  };
+      await mx.publicRooms({ server, limit: 1 });
```
ajbura: i think we should not check for public rooms here(as it delay the input) and adds whatever input as server, user will see the error when they view the server.

src/app/pages/client/explore/Explore.tsx
```diff
@@ -37,40 +41,41 @@ import { AsyncStatus, useAsyncCallback } from '../../../hooks/useAsyncCallback';
...
+      await onSubmit(server);
+      setDialog(false);
```
ajbura: as the function is async it is possible that the component unmount (by user closing the dialog) before the`await` above resolves in that case `setDialog` will crash the application as the component is unmounted. 

you can use `useAlive` hook to conditionally close dialog only if component is `alive()`

ajbura (changes requested):
adding and removing servers works good. but the redesign of explore drawer makes it little confusing. Previous design was good, we can just add the two button in `+ Add server` dialog, one to `Save` (primary) and other to `View` (soft primary).

added server will appear in the same list which include *featured* server, but we can filter featured server so they do not have duplicate from added one.

gingershaped:
> added server will appear in the same list which include _featured_ server, but we can filter featured server so they do not have duplicate from added one.

I think there should be a visual distinction between featured servers and added servers. If the only difference is the absence of the "remove" button, users could become confused as to why they can't remove them.

.vscode/settings.json
gingershaped: You'd think so, but my VSCode instance refuses to recognize the formatter unless it's specifically set like this.

src/app/pages/client/explore/Explore.tsx
gingershaped: I believe the delay is reasonable if it allows typos to be caught early instead of requiring the user to delete the erroneous entry.

gingershaped:
> adding and removing servers works good. but the redesign of explore drawer makes it little confusing.

What do you find confusing about this design? I'd like to see if I can resolve your concerns. In my opinion it's an improvement over the previous design because it has clearer separation between _featured_ servers and rooms which the user cannot remove and _added_ servers which the user has control over. I also prefer having separate "Add server" and "Explore server" buttons, because that allows the "Explore server" button to be highlighted to indicate more clearly to the user that the rooms they're looking at don't belong to a server in the list.

ajbura (commented):
Hey, i feel like the changes in the explore server list layout is making the design confusing. For example:
1. server can be duplicated in featured/added server list.
2. we now have two ways for adding/viewing existing server, increasing the navigation decision complexity. (i feel like we can just have "add server" button opening a dialog with "input" and "save & view" and "view" button, so user can navigate with linear flow without looking for different places.

i feel like keeping it close to the existing design will be more beneficial.

kfiven:
Hi, these changes are completely altering the designed UX we already have in the app. 
1. The left side is for servers and right space is for rooms and spaces, now it's a mix. 
2. The public directory server is unnecessary as that's essentially a server as well and can be added in existing list.
3. The featured and user's own home server should have more visibility as it is now.
4. Server icons are inconsistent (eye, home, bulb, search).

IMHO saving of new searches can be easily achieved by just with the "Bookmark and View" button saving the searched server below the already provided list.

gingershaped:
> IMHO saving of new searches can be easily achieved by just with the "Bookmark and View" button saving the searched server below the already provided list.

Your proposed approach is exactly the one I took originally, but @ajbura expressed dissatisfaction with it. The current UX is my attempt to resolve their criticisms.

kfiven:
I am sorry but all I can see is that they are suggesting staying close to the original design from start:
> Previous design was good, we can just add the two button in + Add server dialog, one to Save (primary) and other to View (soft primary). https://github.com/cinnyapp/cinny/pull/2268#pullrequestreview-2727794788

> i feel like keeping it close to the existing design will be more beneficial. https://github.com/cinnyapp/cinny/pull/2268#pullrequestreview-3170916125

gingershaped:
> I am sorry but all I can see is that they are suggesting staying close to the original design from start:
> > Previous design was good, we can just add the two button in + Add server dialog, one to Save (primary) and other to View (soft primary). https://github.com/cinnyapp/cinny/pull/2268#pullrequestreview-2727794788
> 
> > i feel like keeping it close to the existing design will be more beneficial. https://github.com/cinnyapp/cinny/pull/2268#pullrequestreview-3170916125

I will revert my changes to my earlier design.

ajbura:
Please resolve the merge conflicts.

src/app/hooks/router/useExploreSelected.ts
```diff
@@ -11,7 +11,7 @@ export const useExploreSelected = (): boolean => {
   return !!match;
 };
 
-export const useExploreFeaturedSelected = (): boolean => {
+export const useExploreFeaturedRooms = (): boolean => {
```
ajbura: Can you please revert it? all our hook around navigation follow the existing strategy.

src/app/hooks/useExploreServers.ts
```diff
@@ -0,0 +1,44 @@
...
+import { AccountDataEvent } from '../../types/matrix/accountData';
+import { useAccountData } from './useAccountData';
+import { useMatrixClient } from './useMatrixClient';
+
+export type InCinnyExploreServersContent = {
+  servers?: string[];
+};
+
+export const useExploreServers = (
+  exclude?: string[]
```
ajbura: it will be good idea to exclude from the output of function rather than passing a list to it because currently `addServer` will commit the list which exclude the passes servers.

src/app/pages/client/explore/Explore.tsx
```diff
@@ -93,79 +109,181 @@ export function AddServer() {
                 variant="Surface"
                 size="500"
               >
-                <Box grow="Yes">
-                  <Text size="H4">Add Server</Text>
-                </Box>
+                <Box grow="Yes">{header}</Box>
```
ajbura: not sure why thing change is required?

src/app/pages/client/explore/Explore.tsx
```diff
@@ -93,79 +109,181 @@ export function AddServer() {
...
-      >
-        <Text size="B300" truncate>
-          Add Server
-        </Text>
-      </Button>
+      <NavItem variant="Background" aria-selected={selected}>
+        <NavButton onClick={() => setDialog(true)}>
+          <NavItemContent>{children}</NavItemContent>
+        </NavButton>
+      </NavItem>
```
ajbura: adding `NavItem` inside `AddExploreServerPrompt` is wrong. The nav item should set a flag outside and prompt should be rendered base on that.

src/app/pages/client/explore/Explore.tsx
```diff
@@ -93,79 +109,181 @@ export function AddServer() {
...
-                  {/* <Button
+                  <Button
                     type="submit"
+                    onClick={handleView}
                     variant="Secondary"
+                    fill="Soft"
                     before={
-                      exploreState.status === AsyncStatus.Loading ? (
+                      viewState.status === AsyncStatus.Loading && (
                         <Spinner fill="Solid" variant="Secondary" size="200" />
```
ajbura: button has "Soft" fill but spinner has "Solid". it should have same fill as button it is in.

src/app/pages/client/explore/Explore.tsx
```diff
@@ -93,79 +109,181 @@ export function AddServer() {
...
+                  {viewState.status === AsyncStatus.Error && (
                     <Text style={{ color: color.Critical.Main }} size="T300">
                       Failed to load public rooms. Please try again.
                     </Text>
                   )}
                 </Box>
                 <Box direction="Column" gap="200">
-                  {/* <Button
+                  <Button
                     type="submit"
```
ajbura: Primary button should be of fill="Solid" and "Primary"

src/app/pages/client/explore/Explore.tsx
```diff
@@ -93,79 +109,181 @@ export function AddServer() {
...
                     <Text size="B400">View</Text>
                   </Button>
+                  <Button
+                    type="submit"
+                    onClick={handleSaveView}
+                    variant="Primary"
+                    fill="Soft"
+                    before={
+                      saveViewState.status === AsyncStatus.Loading && (
+                        <Spinner fill="Solid" variant="Secondary" size="200" />
```
ajbura: wrong fill here also

src/app/pages/client/explore/Explore.tsx
```diff
@@ -93,79 +109,181 @@ export function AddServer() {
...
+      </NavItem>
     </>
   );
 }
 
+type ExploreServerNavItemAction = {
+  onClick: () => Promise<void>;
+  icon: IconSrc;
+  alwaysVisible: boolean;
+};
```
ajbura: passing subcomponent props through parent component is wrong design. The entire component should have been passed.

src/app/pages/client/explore/Explore.tsx
```diff
@@ -199,67 +317,68 @@ export function Explore() {
...
-                        </Text>
-                      </Box>
-                    </Box>
-                  </NavItemContent>
-                </NavLink>
-              </NavItem>
+              <ExploreServerNavItem
+                server={userServer}
+                selected={userServer === selectedServer}
+                icon={Icons.Home}
```
ajbura: The icon should be `Server`

src/app/pages/client/explore/Explore.tsx
```diff
@@ -199,67 +317,68 @@ export function Explore() {
...
+              <ExploreServerNavItem
+                server={selectedServer}
+                selected
+                icon={Icons.Server}
+                action={{
+                  alwaysVisible: true,
+                  icon: Icons.Plus,
+                  onClick: () => addServerCallback(selectedServer, true),
+                }}
+              />
```
ajbura: it should be rendered just above the "add server" button so it doesn't switch places on "Plus" "minus" click

src/app/pages/client/explore/Explore.tsx
```diff
@@ -199,67 +317,68 @@ export function Explore() {
...
+            )}
+            {exploringUnlistedServer && selectedServer !== undefined && (
+              <ExploreServerNavItem
+                server={selectedServer}
+                selected
+                icon={Icons.Server}
+                action={{
+                  alwaysVisible: true,
+                  icon: Icons.Plus,
```
ajbura: Bookmark icon will fit more perfectly here ig.

src/app/pages/client/explore/Server.tsx
```diff
@@ -468,6 +476,20 @@ export function PublicRooms() {
     explore({ instance: instanceId, since: undefined });
   };
 
+  const handleOpenMenu: MouseEventHandler<HTMLButtonElement> = (evt) => {
+    setMenuAnchor(evt.currentTarget.getBoundingClientRect());
+  };
```
ajbura, line 481: I think we should remove this menu from here and make save/bookmark button always visible on sidebar. because it is not much discoverable.

ajbura (changes requested):

src/app/pages/client/explore/Explore.tsx
ajbura: by current way user is not able to change server unless the request fulfill. Thats why previous way was choosen.

src/app/pages/client/explore/Server.tsx
gingershaped: @ajbura I added this menu for the sake of mobile users, since they can't hover over a server name to make the remove button appear. Having the button only appear on hover mirrors the behavior of the triple-dot menu in room lists.

trymeouteh:
There is an Add Server button but it does not work as it will only view the rooms and spaces from thst server, not add it to the server list

PR #2305, opened 2025-04-09
add a simple update script

mininmobile:
this is a simple script i wrote for myself to automatically download & unpack the latest cinny version from github releases. it preserves the deployed configuration, and also injects a custom stylesheet (if a `custom.css` file is present next)

unfortunately it relies on `jq` but i still think this script could be useful to some people!


kfiven:
Hi, this seems to be out of scope as it should be upto the person hosting cinny on how they want to update it. Moreover injecting custom css is outside of the project roadmap.

PR #2344, opened 2025-05-28
Add setting to enable 24-hour time format

GimleLarpes:
### Description
Based on #1739.
This PR adds a toggle in the settings page that allows the user to switch between 12-hour and 24-hour time formats, as requested in #60. Major changes from #1739 are that `Time` does not subscribe to `settingsAtom`, and in general subscriptions to `settingsAtom` are minimized as long as code structure doesn't suffer.

This partially but not fully fixes #60

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #1739, opened 2024-05-14
Add setting to enable 24-hour time format

agmm:
### Description

Support for the 24-hour time format has been a frequently requested feature in the Cinny Matrix room and has been mentioned in multiple issues: #60, #1333, #1549, #1648. This PR adds a toggle in the settings page that allows the user to switch between 12-hour and 24-hour time formats.

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #2347, opened 2025-05-31
Add settings to enable 24-hour time format and customizable date format

### LINKED ISSUE #60: Configurable time and date format
Date and time format is hard coded.

Currently the defaults seem to be American English which is fine but having the option to at least set other common formats used around the world even without full localization/internationalization would be a welcome addition.

GimleLarpes:
### Description

Based on #1739  and #2344
Options for configuring date and time formats have long been requested (see #60 and duplicates). This PR aims to add this functionality.

#### This PR adds:
- A "Date & Time" subsection in General settings
- A toggle that allows the user to switch between 12-hour and 24-hour time formats
- A dropdown/text input field to let the user select a date format

As requested in #60.

In addition to the current `D MMM YYYY` format the following options are added: `DD/MM/YYYY`, `MM/DD/YYYY`, `YYYY/MM/DD`, `Custom` - where the user can enter a 16-character dayjs format string.

#### Example image:
image
<br />

Fixes #60 

#### Type of change

- [o] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) (new function params)

### Checklist:


GimleLarpes:
Could this get reviewed @ajbura ?

src/app/features/settings/general/General.tsx
```diff
@@ -341,6 +345,224 @@ function Appearance() {
   );
 }
 
+type CustomDateFormatProps = {
+  dateFormatString: string;
+  setDateFormatString: (format: string) => void;
```
ajbura: can we rename these to `value` and `onChange`? as this component is all about custom date format, naming props the same makes it little hard to read the internals of component.

src/app/features/settings/general/General.tsx
```diff
@@ -341,6 +345,224 @@ function Appearance() {
...
+          <Text size="B400">Save</Text>
+        </Button>
+      </Box>
+    </SettingTile>
+  );
+}
+
+type PresetDateFormatProps = {
+  dateFormatPreset: string;
+  handlePresetChange: (format: string) => void;
```
ajbura: these can also be renamed to `value` and `onChange`

src/app/features/settings/general/General.tsx
```diff
@@ -341,6 +345,224 @@ function Appearance() {
...
+              <Box direction="Column" gap="100" style={{ padding: config.space.S100 }}>
+                {dateFormatItems.map((item) => (
+                  <MenuItem
+                    key={item.format}
+                    size="300"
+                    variant={dateFormatPreset === item.format ? 'Primary' : 'Surface'}
+                    radii="300"
+                    onClick={() => handleSelect(item.format)}
+                  >
+                    <Text size="T300">{item.name}</Text>
```
ajbura: i was wondering will it be a good idea to render the format here instead of format literals? and we do not need it render it in description unless "Custom" option is selected, and the change will be more clear to user.

src/app/features/settings/general/General.tsx
```diff
@@ -341,6 +345,224 @@ function Appearance() {
...
+          </FocusTrap>
+        }
+      />
+    </>
+  );
+}
+
+function SelectDateFormat() {
+  const [dateFormatString, setDateFormatString] = useSetting(settingsAtom, 'dateFormatString');
+  const [dateFormatPreset, setDateFormatPreset] = useState(dateFormatString);
```
ajbura: we can rename this to `selectedDateFormat` and `setSelectedDateFormat`

ajbura (changes requested):
Overall looks really solid!

src/app/features/settings/general/General.tsx
```diff
@@ -341,6 +345,224 @@ function Appearance() {
...
+  };
+
+  const hasChanges = dateFormatCustom !== dateFormatString;
+  return (
+    <SettingTile
+      description={
+        <Text size="T200">
+          View formatting syntax on{' '}
+          <a
+            href="https://day.js.org/docs/en/display/format"
```
kfiven: IMHO we should remove the custom option as this makes dayjs a hard dependency and links to 3rd party sites in interface should be avoided. Also it’s hard to customise for an average user instead we could provide more pre-configured options if there are any other frequently used formats.

kfiven (changes requested):

src/app/features/settings/general/General.tsx
GimleLarpes: AFAIK dayjs is already a hard dependancy, as the pre-configured options are also dayjs strings? There are so many possible preferences regarding things like seperators (ex. / - . : | space (even nothing in some languages)) that I don't think it would be feasible to provide a comparable amount of choice. 
The ability to set a custom date format was also specifically requested and seemingly well received over on [#60](https://github.com/cinnyapp/cinny/issues/60), so I don't know. I personally of course would like to have the functionality, and I don't think it's that difficult to customize since there are basically "templates" included. But I do see the concern with linking to external websites.

We could of course include more options, I just think it's cleaner to have fewer options in the dropdown.

GimleLarpes: AFAIK dayjs is already a hard dependancy, as the pre-configured options are also dayjs strings? There are so many possible preferences regarding things like seperators (ex. `/` `-` `.` `:` ` ` (even nothing in some languages)) that I don't think it would be feasible to provide a comparable amount of choice.
The ability to set a custom date format was also specifically requested and seemingly well received over on #60 , so I don't know. I personally of course would like to have the functionality, and I don't think it's that difficult to customize since there are basically "templates" included. But I do see the concern with linking to external websites.

We could of course include more options as well, I just think it's cleaner to have fewer options in the dropdown.

kfiven: Hard dependency meant, we can replace the pre-configured options in the code and have same functionality but not the custom ones as they will be saved in local storage and could be in any format.

In case of separator, I think `-` is the only another widely used so we can add one option for that.

GimleLarpes: I see, is removing dayjs being considered? If so, then yeah, this would at least have to wait until the swap has been made. 
If not though I would like to give users extra functionality now and then in the worst case remove it later, instead of never giving it since there is a possibility it might have to be removed / updated in the future.

In the English speaking world `-` might be the only other widely used one, but the standard in my region is `.`, and quite a few seem to use `. ` and characters/kanji as standard as well, not taking individual preference into account. Adding more settings to get some customization back also feels very clunky compared to just letting people choose.

ajbura: i also feel like having dayjs doc link is wrong direction. can we think of any other options about it?

GimleLarpes: We could maybe have a "help" page/popup that shows the syntax without going to external websites? I know that people have been asking for a page like that for commands, so it might get reused for that as well.

ajbura: maybe we can have a info menu in custom input just like we have in login screen for username field.

ajbura: @GimleLarpes pinging in case you missed my feedback.^^^^^

GimleLarpes: Good idea, I'll try doing something similar to the info menu.

GimleLarpes:
I've now added a hint for the formatting. I would've preferred to prevent overflow using `flexFlow: 'column wrap'` or something similar, but the current method works. I've also added 24hr functionality to the TimePicker.

ajbura (approved):
Looks Great! Thanks for such a nice addition.

kfiven (approved):

PR #2349, opened 2025-06-01
Fix focus behaviour when opening single-purpose pages, and rooms

### LINKED ISSUE #2261: Auto select text input
### Describe the problem

When clicking the "Join with Address" button, I need to click the input field myself. 

I would like the field to be selected automatically. 

Image

Image

### Alternatives considered

_No response_

### Additional context

_No response_

### LINKED ISSUE #2348: Improve focus behaviour on single-purpose pages/popups and room message field
This is an extension of issue #2261.

### Describe the problem
Currently only the `CTRL+K` search bar gains focus automatically when it is opened. This means that you have to do extra inputs before being able to use many features such as searching messages when clicking `"Message Search"`, or entering the user to DM when clicking `"Create Chat"`. This adds extra friction to using the app and degrades the user experience, especially in daily use and on mobile devices where clicking is more finicky.

In addition, on desktop it is not clear that the user does not need to click the message field to begin typing. Therefore I suggest making this clear by focusing the message field when opening rooms on non-mobile devices.

I would like the focus behavior to change so that,
The following always gain focus:
- "Join with Address"
- "Message Search" (Spaces and Home)
- "Create Chat" (new DM, invite to room)

The following gain focus when not on mobile:
- Room message field

Components mentioned are `"Message Search"`, `"Create Chat"`, `"Invite"`, `"Join with Address"`, and the room message field.

GimleLarpes:
### Description

This PR improves focus behaviour as formulated in #2348  when opening pages/popups that have a singular purpose.
It also focuses the message field when opening a room, to more clearly indicate to the user that they can begin typing immediately. 

The reasoning for suggesting these changes is primarily that pages where you can only do one thing, should let you do those things in one click (ie. search, creating new DM). Focusing on the message field enables simply entering a room and immediately starting to type (while the user can currently do this, that is not communicated to the user). Not gaining focus on mobile prevents the onscreen keyboard from appearing when it's not wanted (ie. when just reading messages).

This also unifies the UX as `CTRL+K` search already gains focus automatically.<br><br>

#### This PR makes the following changes:
The following always gain focus:
- "Join with Address"
- "Message Search" (Spaces and Home)
- "Create Chat" (new DM, invite to room)

The following gain focus when not on mobile:
- Room message field<br><br>

#### Example videos:
Current:

https://github.com/user-attachments/assets/558c55a6-6ba6-44df-a900-579b5d837b21

https://github.com/user-attachments/assets/71c01e1b-69c6-4649-9099-de8b59a0b19d

After changes:

https://github.com/user-attachments/assets/44f5c9a4-4fcb-4b61-a80b-c8f9adae39f9

https://github.com/user-attachments/assets/3595fb04-0fad-46ef-af8d-41ac57d0f27f

Fixes #2348, Fixes #2261 

#### Type of change

### Checklist:

### DIFF
--- a/src/app/features/room/RoomView.tsx
+++ b/src/app/features/room/RoomView.tsx
@@ -1,9 +1,10 @@
-import React, { useCallback, useRef } from 'react';
+import React, { useCallback, useEffect, useRef } from 'react';
 import { Box, Text, config } from 'folds';
 import { EventType, Room } from 'matrix-js-sdk';
 import { ReactEditor } from 'slate-react';
 import { isKeyHotkey } from 'is-hotkey';
 import { useStateEvent } from '../../hooks/useStateEvent';
+import { ScreenSize, useScreenSizeContext } from '../../hooks/useScreenSize';
 import { StateEvent } from '../../../types/matrix/room';
 import { usePowerLevelsAPI, usePowerLevelsContext } from '../../hooks/usePowerLevels';
 import { useMatrixClient } from '../../hooks/useMatrixClient';
@@ -63,6 +64,8 @@ export function RoomView({ room, eventId }: { room: Room; eventId?: string }) {
 
   const [hideActivity] = useSetting(settingsAtom, 'hideActivity');
 
+  const screenSize = useScreenSizeContext();
+
   const { roomId } = room;
   const editor = useEditor();
 
@@ -98,6 +101,12 @@ export function RoomView({ room, eventId }: { room: Room; eventId?: string }) {
       [editor]
     )
   );
+
+  useEffect(() => {
+    if (screenSize === ScreenSize.Desktop) {
+      ReactEditor.focus(editor);
+    }
+  }, [editor, screenSize]);
 
   return (
     <Page ref={roomViewRef}>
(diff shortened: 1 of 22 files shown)


src/app/features/room/RoomView.tsx
```diff
@@ -99,6 +101,12 @@ export function RoomView({ room, eventId }: { room: Room; eventId?: string }) {
     )
   );
 
+  useEffect(() => {
+    if (screenSize === ScreenSize.Desktop && canMessage && editor) {
+      ReactEditor.focus(editor);
+    }
+  }, [editor, screenSize, canMessage]);
+
```
ajbura: This change can be problematic on Tablet device, most of which are wide enough to be detected as Desktop *like* size. Problem will be auto virtual keyboard appearing on every room switch.

ajbura (changes requested):

src/app/features/room/RoomView.tsx
GimleLarpes: I see, do you know of a better way to do it? I'll try to find a decent workaround / test it on some tablets but I might drop it if nothing is found.

GimleLarpes: Immediately the best way seems to be adding `&& !isTouch `, where `const isTouch = navigator.maxTouchPoints > 0;`. However this would disable focusing on touch screen laptops as well.

Otherwise, tablets should ideally be caught by ScreenSize.Tablet. I tested on an iPad 8th gen which has a resolution of 1620x2160 (a lot more than the coded max tablet width of 1124), and it worked. Is there a tablet that we know ScreenSize.Tablet doesn't work for?

ajbura: I think it would be safe to drop this change as user can already type to focus.

ajbura (approved):

PR #2361, opened 2025-06-12
Add code block copy and collapse functionality

### LINKED ISSUE #1110: Allow to collapse some elements such as big code block
### Describe the problem

Sometimes someone or a bot might post some big blocks such as block code, or quote a particularly long issue, which might take your whole screen and even more.

It would come handy to be able to have this collapsed (as done in Element, for example)

### Alternatives considered

None

### Additional context

_No response_

### LINKED ISSUE #1592: Add a button to copy code block
### Describe the problem

Sometimes I'd like to copy a codeblock and it's generally inconvenient to try to select everything manually.

Add a button to copy the entire codeblock to clipboard. Something similar to discord would work:

image

### Alternatives considered

_No response_

### Additional context

_No response_

GimleLarpes:
### Description

The ability to copy and collapse code blocks have been requested in #1110 and #1592. This PR aims to add this functionality in a similar way to Element web and Discord.

This PR makes it so that:
- Large code blocks (currently more than 14 lines) are collapsed into a scrollable 10-line preview that can be expanded to the full message length
- Adds a copy button that copies the contents of the code block
- Adds a collapse/expand button to code blocks

This prevents large code blocks from clogging up conversations.<br></br>

#### Example Images
Collapsed, non-hovered
image

Expanded, hovered
image

Fixes #1110, Fixes #1592

#### Type of change

### Checklist:


src/app/plugins/react-custom-html-parser.tsx
```diff
@@ -195,6 +204,114 @@ export const highlightText = (
...
+          variant="Secondary"
+          size="300"
+          radii="300"
+          onClick={onToggle}
+          aria-expanded={!collapsed}
+          aria-controls="code-block-content"
+          aria-label={collapsed ? 'Show Full Code Block' : 'Show Code Block Preview'}
+          style={collapsed ? { visibility: 'visible' } : {}}
+        >
+          <Icon src={collapsed ? Icons.ChevronRight : Icons.ChevronBottom} size="50" />
```
ajbura: The correct icon here would be `collapsed ? ...Bottom : ...Top` as the icon-button should represent action rather than state.

src/app/plugins/react-custom-html-parser.tsx
```diff
@@ -195,6 +204,114 @@ export const highlightText = (
...
+      >
+        <Icon src={copied ? Icons.Check : Icons.File} size="50" />
+      </IconButton>
+      {collapsible && (
+        <IconButton
+          variant="Secondary"
+          size="300"
+          radii="300"
+          onClick={onToggle}
+          aria-expanded={!collapsed}
```
ajbura: we an use aria-pressed so icon can have the visual clue of active state.

src/app/plugins/react-custom-html-parser.tsx
```diff
@@ -195,6 +204,114 @@ export const highlightText = (
...
+    // Needs a better copy icon
+    <div className={css.CodeBlockControls}>
+      <IconButton
+        variant="Secondary"
+        size="300"
+        radii="300"
+        onClick={onCopy}
+        aria-label="Copy Code Block"
+      >
+        <Icon src={copied ? Icons.Check : Icons.File} size="50" />
```
ajbura: after clicking copy the icon should not left with the `Check` icon and should reset back to previous state after like half a second. it would be helpful to have a generic hook as we really need this functionality at various other place3s.

src/app/plugins/react-custom-html-parser.tsx
```diff
@@ -195,6 +204,114 @@ export const highlightText = (
...
+
+  return (
+    <>
+      <CodeBlockControls
+        copied={copied}
+        onCopy={handleCopy}
+        collapsible={collapsible}
+        collapsed={collapsed}
+        onToggle={toggleCollapse}
+      />
```
ajbura: do we really need a component for that if it only just has other components without any other functionality? it looks like we are going into prop drilling without need.

ajbura: these controls hides the code under them if the first line is long. as i feel like adding the padding top in code view look bad, will it be a good idea if we can move them to bottom of code block without hover effect (which doesn't work without mouse or touch)? we can also consider proper `Chip` buttons with text making it more actions more readable.

src/app/styles/CustomHtml.css.ts
```diff
@@ -85,10 +85,35 @@ export const CodeBlock = style([
...
     fontStyle: 'normal',
+    position: 'relative',
   },
 ]);
-export const CodeBlockInternal = style({
-  padding: `${config.space.S200} ${config.space.S200} 0`,
+export const CodeBlockInternal = recipe({
+  base: {
+    padding: `${config.space.S200} ${config.space.S200} 0`,
+    minWidth: config.size.ModalDrawerWidth,
```
ajbura: the uses of `ModalDrawerWidth` is incorrect, you can use the `toRem` function to have a custom width.

src/app/styles/CustomHtml.css.ts
```diff
@@ -85,10 +85,35 @@ export const CodeBlock = style([
...
+  },
+  variants: {
+    collapsed: {
+      true: {
+        maxHeight: `calc(${config.size.ModalDrawerWidth} - ${config.lineHeight.T400} / 2)`,
```
ajbura: it is also not super clear to me whats happening here. i feel you want to have a square shape when codeblock is collapsed?  adding more explicit value would be better ig.

ajbura (changes requested):

src/app/plugins/react-custom-html-parser.tsx
GimleLarpes: My thinking here was that it would look the same as the icon in space category headers to avoid confusion (i.e. collapsed things are associated with the same icon everywhere in the UI). I do see your point though and I'll ofc change it if you want me to.

src/app/styles/CustomHtml.css.ts
GimleLarpes: The math was due to `ModalDrawerWidth `being exactly 10 lines, which didn't indicate that the block could be expanded. The best option is maybe to simply use `9.5 * ${config.lineHeight.T400}`?

src/app/plugins/react-custom-html-parser.tsx
GimleLarpes: I feel moving the collapse button to the bottom might be confusing (at least I expect it to be at the top), but I can test it out.

Also, clarification: Do you mean to change the absolute positioning to be at the bottom instead of top (this has the same issue as currently, but for the last line of code), or to move it after the code block (column layout)?

ajbura: No you can move it outside of the `Scroll` or even better inside the `Scroll` but with the position as sticky "bottom", so it will stay floating at bottom but at the end do not hide the line under it.

src/app/plugins/react-custom-html-parser.tsx
ajbura: Yeah so space category headers are "header" which show state of the section i.e "open/close state + name". they are not suppose to be clickable but we made them for convenience so we do not have to add a new button for that. otherwise in all buttons we have bottom icon to expand section/view

src/app/styles/CustomHtml.css.ts
ajbura: that is why i think placing the expand button at the bottom will be helpful. it will prompt the user to do the action compare to finding that they have a truncated section and later finding option to expand the section.

src/app/plugins/react-custom-html-parser.tsx
GimleLarpes: I've implemented what I think you mean, but I don't think it works very well as it takes a lot of space (especially on 1-line blocks) and you cannot "peek into" a code block by quickly expanding and collapsing the code block, since the expand button jumps around and may disappear from view when expanding and collapsing.

The half-line of text also looks a bit weird when there's seemingly open space underneath it.  But if you remove the half-covered line it' simply looks like the code block ends.
image

I believe the best solution would be something based on the original idea. The problem with the expand button covering the first line of code while possibly annoying is quite minor, as expanding the block will let you see everything. The biggest issue is accessibility as the copy code button is a bit clunky to get to.

src/app/plugins/react-custom-html-parser.tsx
```diff
@@ -195,6 +205,88 @@ export const highlightText = (
...
+            aria-controls="code-block-content"
+            aria-label={collapsed ? 'Show Full Code Block' : 'Show Code Block Preview'}
+            style={collapsed ? { visibility: 'visible' } : {}}
+          >
+            <Icon src={collapsed ? Icons.ChevronBottom : Icons.ChevronTop} size="50" />
+          </IconButton>
+        )}
+      </div>
+      <Scroll
+        direction={collapsed ? 'Both' : 'Horizontal'}
```
ajbura: when user open the codeblock it hide the vertical scrollbar which makes the codeblock little smaller is in width, which result in a content shift like effect. it will be good if we can prevent that.

ajbura (changes requested):
thank you for trying out different ideas, i feel like it is good to go for now except few minor changes. 

I will add the new copy button later and will try what can be done to prevent the "does codeblock is scrollable" issue.

ajbura (approved):
LGTM. Thanks

PR #2335, opened 2025-05-23
Adds Element Call support, call UI elements, and rudimentary widget support

GigiaJ:
### Description
Adds Element Call support for both the npm embedded package as well as a hosted instance.
Adds a level of rudimentary widget support (not entirely sure of the extent even needed).
Adds calling interface elements such as a nav bar status section with a mute/unmute button, video enable/disable button, and hang up button. Icons are somewhat of a placeholder, but are relatively fitting.
Added a segmentation between Text rooms and Call rooms and provided an icon for call rooms (need to add the visibility icon support such as public/private/etc)
Added a --disabled for now-- icon for calling in DMs in the room header
Added dual iframe state juggling to provide lobby support akin to Discord where a user might wish to view another call's lobby or text channel
Added a text channel toggle to show or hide the room text channel
Added user icons and usernames to appear in the room nav to provide call activity visibility akin to Discord's previous interface implementation
Added mobile support for calling and text channel swapping and made the default behavior on mobile simply place you in the lobby and never in a call (does currently pop on your camera and mic, but is not joined to a call.)
Adjusted Room Header to account for a call room vs text room and behave accordingly

** Added a call or text room creation option in the room creation menu
Probably some other bits I'm forgetting, apologies. I feel like this is at a good enough stage to open a dialogue and move forward with, so I'm opening a PR.

Fixes #
#528

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas (I did, but comments didn't match the code style of the project. Happy to retroactively add them though.)
- [ ] I have made corresponding changes to the documentation (I'd happily add docs pertaining to calling or simply hosting the standalone element call and SFU)
- [ ] My changes generate no new warnings (The iframes probably will always generate some level as widgets are created and destroyed)

PR #2367, opened 2025-06-19
Fix room header members icon not filled when enabled

GimleLarpes:
### Description

The current behaviour of the "Members" button in room headers is incorrect, as it is not filled when the members drawer is open.

This PR makes it so that the "Members" button is filled when enabled, as is the correct behaviour according to [[1997kB]](https://matrix.to/#/#cinnyapp:matrix.org/$SEY6VlFajDU9NaUQz2tqtqJLZjuBeUBJMhLtQAH-t9c).

Fixes #

#### Type of change

### Checklist:

### DIFF
--- a/src/app/features/room/RoomViewHeader.tsx
+++ b/src/app/features/room/RoomViewHeader.tsx
@@ -34,7 +34,7 @@ import { RoomTopicViewer } from '../../components/room-topic-viewer';
 import { StateEvent } from '../../../types/matrix/room';
 import { useMatrixClient } from '../../hooks/useMatrixClient';
 import { useRoom } from '../../hooks/useRoom';
-import { useSetSetting, useSetting } from '../../state/hooks/settings';
+import { useSetting } from '../../state/hooks/settings';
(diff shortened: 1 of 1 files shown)


ajbura:
The "Members" icon is not filled on click because the functionality it toggle is a view that is detached from current toggle unlike something like "Pinned Messages" (where solid icon show open state prominently). It act as preference toggle to view/hide the other side view and choice of keeping it open can't be show as prominent choice.

GimleLarpes:
I see, so the Members icon should be permanently non-filled. Would the same apply to the "Chat" button in #2335 ? (see image)
image

ajbura:
> I see, so the Members icon should be permanently non-filled. Would the same apply to the "Chat" button in #2335 ?

Well most likely yes, if the chat view will open as sidepanel but as in above screenshot the chatview appears to be under room header occupying the main area of callview... it a bit challenging to know how they are approaching the feature

GimleLarpes:
Got it, I'll close this PR now.

PR #2366, opened 2025-06-19
Mobile UI context menus and slide to reply and edit (relies on #2335 #2345 #2356)

GigiaJ:
### Description
RELIES ON #2335 #2356 #2345
Should not be merged before those. (My fault, but I would be creating merge conflicts between my own changes unnecessarily otherwise.)

Adds mobile gesture context menus with a reusable component for adding more in the future. These menus additionally can be dragged down to closed. 
Added slide to reply and long slide to edit for individual messages.

Fixes #
#17 #1647
#### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [o] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #2369, opened 2025-06-21
Use the modern multi-account session handling (and transition old sessions in)

GigiaJ:
### Description
Important, but this should be merged after #2366 and the requirements it has.
It isn't a hard requirement though and the merge conflicts relatively will be mild.

Adds the multi-account session handling in its entirety and removes legacy references.
Shouldn't impact users at all, but allows me to move forward with better interfacing with the local storage. (Suffering a good number of issues with asynchronously doing so as well as session states fail after refresh on modifying it. This solves that.)

No actual interface for multiple accounts, but in theory they should be doable now.

Fixes #

#### Type of change

### Checklist:

### DIFF
--- a/src/app/state/sessions.ts
+++ b/src/app/state/sessions.ts
@@ -1,4 +1,3 @@
-import { atom } from 'jotai';
 import {
   atomWithLocalStorage,
   getLocalStorageItem,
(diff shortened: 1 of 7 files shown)


kfiven:
We do not have this in roadmap yet. Will revisit once we move over to new SDK.

PR #2377, opened 2025-06-26
Jump to time

ajbura:
### Description

Fixes #

#### Type of change

### Checklist:

### DIFF
--- a/src/app/features/room/RoomViewHeader.tsx
+++ b/src/app/features/room/RoomViewHeader.tsx
@@ -65,6 +65,8 @@ import {
   getRoomNotificationModeIcon,
   useRoomsNotificationPreferencesContext,
 } from '../../hooks/useRoomsNotificationPreferences';
+import { JumpToTime } from './jump-to-time';
+import { useRoomNavigate } from '../../hooks/useRoomNavigate';
 
 type RoomMenuProps = {
   room: Room;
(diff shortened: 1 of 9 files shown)


src/app/features/room/RoomViewHeader.tsx
GigiaJ, line 1: I feel like this makes more sense existing in the message search page and being able to search for messages between a certain range and jump to them from there in addition to being able to apply filters.

ajbura: we do not have any server side endpoint that can combine these functionality. These are both two diff functionality given by spec as two diff endpoints.

GigiaJ: Not disagreeing given that the options client side are terrible seeing as you'd have to mash /timestamp_to_event, /search, /messages, and potentially /context to create a what would effectively be a glorified super inefficient SQL join. Server side, it just looks like an issue is opened, but no proposal.

That said, I imagine that ideally the goal is to keep the room header uncluttered and search-like behaviors visually associated. Apologies though, I should've kept my actual message clear and concise instead of throwing in wants. Haha.

kfiven: This is more of navigating the room timeline rather than search option, so in my opinion it makes to have it in room header. Though I also agree we should not get this out of “Options” (from cluttering prospective).

kfiven (approved):

PR #2395, opened 2025-07-15
Update presence status when the window loses focus.

krishukr:
(cherry picked from commit 9c99a84ad0513abf33c119ab68590997435f62c8)

### Description

Background: My friends complained that I always show online but don’t read messages in time and confused. I discovered that Cinny keeps setting my account’s presence status to `online` even when I put the client in the background.

Apparently, Cinny sent `/sync` requests to Homeserver even in the background, which is reasonable for a client to catch up with events. However, this misled the HS to believe that the user was `online`. 

Investigating Matrix Client-Server API specification, I discovered that an user's presence status is inferred by Homeserver based on API activities or be explicitly set by Client via [`/presence/{userId}/status`](https://spec.matrix.org/v1.15/client-server-api/#put_matrixclientv3presenceuseridstatus) or implicitly set along with [`/sync`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync) requests using parameter `set_presence`. `matrix-js-sdk` provides both methods [`setPresence`](https://matrix-org.github.io/matrix-js-sdk/classes/matrix.MatrixClient.html#setpresence) and [`setSyncPresence`](https://matrix-org.github.io/matrix-js-sdk/classes/matrix.MatrixClient.html#setsyncpresence).

Ideally, we choose to use `setSyncPresence`. Because even if presence status were set via `/presence`, `/sync` API calls that keep running in the background still confuse Homeserver to consider the user being `online`. Furthermore, this way won't override possible presence status from other clients.

After this change, a) when the Cinny window is focused, the presence is `online`, b) when the Cinny window is *not* focused and running in background, the presence is `unavailable`, c) when the Cinny isn't running, the presence is `offline`. 

#### Type of change

### Checklist:

### DIFF
--- a/src/app/pages/client/ClientNonUIFeatures.tsx
+++ b/src/app/pages/client/ClientNonUIFeatures.tsx
@@ -1,7 +1,7 @@
 import { useAtomValue } from 'jotai';
 import React, { ReactNode, useCallback, useEffect, useRef } from 'react';
 import { useNavigate } from 'react-router-dom';
-import { RoomEvent, RoomEventHandlerMap } from 'matrix-js-sdk';
+import { RoomEvent, RoomEventHandlerMap, SetPresence } from 'matrix-js-sdk';
 import { roomToUnreadAtom, unreadEqual, unreadInfoToUnread } from '../../state/room/roomToUnread';
 import LogoSVG from '../../../../public/res/svg/cinny.svg';
 import LogoUnreadSVG from '../../../../public/res/svg/cinny-unread.svg';
@@ -253,6 +253,28 @@ function MessageNotifications() {
   );
 }
 
+function UpdatePresence() {
+  const mx = useMatrixClient();
+  const onFocus = () => {
+    mx.setSyncPresence(SetPresence.Online);
+  };
+  const onBlur = () => {
+    mx.setSyncPresence(SetPresence.Unavailable);
+  }
+
+  useEffect(() => {
+    window.addEventListener("focus", onFocus);
+    window.addEventListener("blur", onBlur);
+    onFocus();
+    return () => {
+      window.removeEventListener("focus", onFocus);
+      window.removeEventListener("blur", onBlur);
+    };
+  }, [mx]);
+
+  return null;
+}
+
 type ClientNonUIFeaturesProps = {
   children: ReactNode;
 };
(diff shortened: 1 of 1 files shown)


krishukr:

ajbura (commented):
Hey, Thanks. The changes mostly look good, but i could not able to test them as my server doesn't have presence feature enabled. Could you recommend any server which does that and have open registration?

Also it might be a good idea to only `setSyncPresence` if server actually support it.

krishukr:
> Hey, Thanks. The changes mostly look good, but i could not able to test them as my server doesn't have presence feature enabled. Could you recommend any server which does that and have open registration?
> 
> Also it might be a good idea to only `setSyncPresence` if server actually support it.

Thanks for responding, I am new to the Matrix Ecosystem and not familiar with it, so please do point it out if I am wrong.

I personally use [Synapse](https://github.com/element-hq/synapse) as my HS. Which, if I recollect correctly, is also the HS for matrix.org. 

As it for reproducing, my guess is that you are using Cinny as your client, which does not display presence of users. You can try [Element](https://element.io/app). 

But the easiest way is to run a curl command in shell: 

```sh
curl -X GET https://<YOUR_HOMESERVER>/_matrix/client/v3/presence/<ANY_USER_YOU_LIKE>/status \
    --header 'Authorization: Bearer <YOUR_ACCESS_TOKEN>'
```

Access token can be get in Cinny's Settings > Developer Tools. Enable Developer Tools first, and copy Access Token. This API is documented in Matrix Client-Server API [spec](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3presenceuseridstatus).

Here I also quote from Matrix [spec](https://spec.matrix.org/v1.15/client-server-api/#modules):
> Modules are strictly defined within this specification and should not be mistaken for experimental extensions or optional features. A compliant server implementation MUST support all modules and supporting specification[.]

Where [Presence](https://spec.matrix.org/v1.15/client-server-api/#presence) defined as a module, most popular matrix server should implemented this module, i.e., Cinny should properly handle presence status accordingly.

Setting presence status along with a `/sync` call is also in [spec](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync).

ajbura:
yeah synapse do support presence but `matrix-client.matrix.org` has this feature disabled so everyone remain 'offline' for me, that is why i am not able test it on matrix.org and wondering if there are any server with open registration for that

PR #2406, opened 2025-07-28
Change: Prevent publishing rooms with incompatible joinrules to directory

GimleLarpes:
### Description

Currently rooms that cannot be joined from the directory can be published to the directory. To fix this I suggest disabling the option to publish if the room has incompatible joinrules such as "Invite Only" or "Space Members Only"

Fixes #

#### Type of change

### Checklist:

- [o] I have commented my code, particularly in hard-to-understand areas
- [o] I have made corresponding changes to the documentation

### DIFF
--- a/src/app/features/common-settings/general/RoomPublish.tsx
+++ b/src/app/features/common-settings/general/RoomPublish.tsx
@@ -12,6 +12,7 @@ import { IPowerLevels, powerLevelAPI } from '../../../hooks/usePowerLevels';
 import { StateEvent } from '../../../../types/matrix/room';
 import { useMatrixClient } from '../../../hooks/useMatrixClient';
 import { useStateEvent } from '../../../hooks/useStateEvent';
+import { ExtendedJoinRules } from '../../../components/JoinRulesSwitcher';
 
 type RoomPublishProps = {
   powerLevels: IPowerLevels;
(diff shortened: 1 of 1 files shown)


GimleLarpes:
I do have a question: Why is 

```
  const joinRuleEvent = useStateEvent(room, StateEvent.RoomJoinRules);
  const content = joinRuleEvent?.getContent<RoomJoinRulesEventContent>();
  const rule: JoinRule = content?.join_rule ?? JoinRule.Invite;
```

used instead of `const rule: JoinRule = room.getJoinRule();` in `RoomJoinRules.tsx` to get the joinrule?

ajbura:
> I do have a question: Why is

`room.getJoinRule()` will not do the re-render on join rule change but using it via `useStateEvent` will re-render the component after state event update. we are using it as we want UI to be updated when join rule gets update from other places/client

GimleLarpes:
I see, thank you for the explanation!

src/app/features/common-settings/general/RoomPublish.tsx
```diff
@@ -23,13 +25,17 @@ export function RoomPublish({ powerLevels }: RoomPublishProps) {
...
+  const content = joinRuleEvent?.getContent<RoomJoinRulesEventContent>();
+  const rule: JoinRule = content?.join_rule ?? JoinRule.Invite;
 
   const { visibilityState, setVisibility } = useRoomDirectoryVisibility(room.roomId);
 
   const [toggleState, toggleVisibility] = useAsyncCallback(setVisibility);
 
   const loading =
     visibilityState.status === AsyncStatus.Loading || toggleState.status === AsyncStatus.Loading;
+  const validRule = rule === JoinRule.Public || rule === JoinRule.Knock;
```
ajbura: we also need to check for `knock_restricted` join rule here which is part of `ExtendedJoinRule` type

ajbura (changes requested):

ajbura (approved):

PR #2455, opened 2025-08-19
feat: allow using filenames in codeblocks

WebFreak001:
### Description

Makes it possible to send code blocks as \`\`\`filename.json or \`\`\`src/app/example.js and having Cinny display the filename in the header of code blocks as well as only using the text after the dot as the language specifier.

- If there is a dot in the syntax name, we instead treat it as the filename and only everything after the last dot as the syntax name
- we use a custom "data-label" attribute on the code block to specify the name of the file (so only compatible with new Cinny versions, no support for other matrix clients and no spec/standard for this)

Preview in Cinny:

image

&nbsp;

Preview in old Cinny and probably other clients:

image

#### Type of change

### Checklist:

### DIFF
--- a/src/app/plugins/markdown/block/rules.ts
+++ b/src/app/plugins/markdown/block/rules.ts
@@ -16,8 +16,12 @@ export const CodeBlockRule: BlockMDRule = {
   match: (text) => text.match(CODEBLOCK_REG_1),
   html: (match) => {
     const [, g1, g2] = match;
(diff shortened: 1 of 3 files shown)


ajbura (commented):
~~While this can work within cinny this will break the every other client code highlight~~

ajbura (dismissed):
While this looks fine it doesn't include changes to retrieve filename when editing a message.

kfiven (approved):

PR #2522, opened 2025-10-28
Add support for rainbow and rainbowme commands and /html

RGBCube:
#### Type of change

### Checklist:

### DIFF
--- a/src/app/features/room/RoomInput.tsx
+++ b/src/app/features/room/RoomInput.tsx
@@ -326,6 +326,8 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
         plainText = rainbow(plainText);
         customHtml = rainbow(customHtml);
         msgType = MsgType.Emote;
+      } else if (commandName === Command.Html) {
+        customHtml = plainText;
       } else if (commandName === Command.Shrug) {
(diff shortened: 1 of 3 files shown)


RGBCube:
Can confirm it works now:

image

Have also tested it with /html for a \<details>, worked well.

RGBCube:
cc @ajbura can we get this in soon? this was a continuation of an older PR and the code has attribution

ajbura:
Hey, @RGBCube currently the rainbow commands renders the raw html if you put any other formatted text in your message. Also if we use the plain text only to generate the rainbow html we will loose the formatting entirely which will create another issue. So I think we probably need to tweak the algorithm so it can process html without breaking it?

RGBCube:
> Hey, @RGBCube currently the rainbow commands renders the raw html if you put any other formatted text in your message. Also if we use the plain text only to generate the rainbow html we will loose the formatting entirely which will create another issue. So I think we probably need to tweak the algorithm so it can process html without breaking it?

Ah, yeah I hadn't thought of that. It would be easy to simply ignore any markdown and send as-is, or we could treat it as a dom and recurse inwards. I think the latter is better, I'll try to tackle it.

PR #2562, opened 2026-01-01
Fix read marker triggering room-wide notifications on alias mention

### LINKED ISSUE #2533: When mentioning a room's own alias in the room, it sends unintended room ping (similar to old @room behavior)
### Describe the bug

When in a room with alias #some-room:example.com you add that alias into the message, it pings all users within the room somehow

### Reproduction

1. Create a room
2. Assign an alias to that room (example: `#some-room:example.com`)
3. Send a message in that room that will have `#some-room:example.com` mentioned somewhere in the text
4. Wake up everyone in that room, similarly to `@room`

### Expected behavior

Mentioning of the room's own alias should not result into room mention (@room-like notification for all room members)

### Platform and versions

```shell
Cinny v4.10.2
```

### Additional context

That happened in the #mrs:etke.cc room, specific event: https://matrix.to/#/#mrs:etke.cc/$NJVpI4sREce4qtywIWPElyKvAR2Qp5_zlb_vHBi9p1c

Below is shrinked json of the event's content:

```
{
  "content": {
    "body": "..., #mrs:etke.cc , ...",
    "format": "org.matrix.custom.html",
    "formatted_body": "..., <a href=\"https://matrix.to/#/#mrs:etke.cc\">#Matrix Rooms Search</a> , ...",
    "m.mentions": {
      "room": true,
      "user_ids": [
        "REDACTED_VALID_MXID_TO_WHOM_REPLY_WAS_ADDRESSED"
      ]
    },
    "m.relates_to": {
      "m.in_reply_to": {
        "event_id": "$RuTC0O7jKZuDcZv-ZKyzBqomi8QZe5m4iw6BmEOD08U"
      }
    },
    "msgtype": "m.text"
  },
}
```

@nexy7574 mentioned they may know more details about the root cause of the issue

Santhoshkumar044:
## Summary
Fixes bug where mentioning a room's alias (e.g., `#room:server.com`) incorrectly triggers `@room`-like notifications for all members instead of treating it as a regular link.

## Problem
When a message contained the current room's alias, the `getMentions()` function incorrectly set `"room": true` in the `m.mentions` field. This caused all members with "Contains @room: Notify Loud" enabled to receive notifications, even when the room alias was mentioned as a regular link and not as an `@room` mention.

**Root cause:** Both `@room` mentions (via autocomplete) and manually typed room aliases resolve to the same room alias ID, making them indistinguishable when only checking the `node.id`.

## Solution
Modified the `getMentions()` function to distinguish between:
- **@room mentions** → Display name doesn't start with `#` → Sets `room: true` ✓
- **Room alias mentions** → Display name starts with `#` → Treated as regular link ✓

The fix checks the `node.name` (display text) instead of just the `node.id` to determine intent.

## Changes
- Updated `getMentions()` in `src/app/components/editor/util.ts` (or your file path)
- Added condition: `if (!node.name.startsWith('#'))` before setting `mentionData.room = true`
- This ensures only actual `@room` mentions trigger room-wide notifications

## Testing
Tested with 3 users in a room with alias `#test:matrix.org`:

**Before fix:**
- ❌ `@user:server.com welcome to #test:matrix.org` → All members notified
- ❌ `Check #test:matrix.org` → All members notified

**After fix:**
- ✅ `@room hello` → All members notified 
- ✅ `@user:server.com welcome to #test:matrix.org` → Only @user notified
- ✅ `Check #test:matrix.org` → No one notified

## Issue Number
Fixes #2533

### DIFF
--- a/src/app/components/editor/output.ts
+++ b/src/app/components/editor/output.ts
@@ -212,11 +212,15 @@ export const getMentions = (mx: MatrixClient, roomId: string, editor: Editor): M
     if (node.type === BlockType.CodeBlock) return;
 
     if (node.type === BlockType.Mention) {
-      if (node.id === getCanonicalAliasOrRoomId(mx, roomId)) {
-        mentionData.room = true;
-      }
-      if (isUserId(node.id) && node.id !== mx.getUserId()) {
-        mentionData.users.add(node.id);
+      const mentionId = node.id;      
+      if (mentionId === getCanonicalAliasOrRoomId(mx, roomId)) {
+        if (!node.name.startsWith('#')) {
+          mentionData.room = true;
+        } 
+      } 
+      
+      if (isUserId(mentionId) && mentionId !== mx.getUserId()) {
+        mentionData.users.add(mentionId);
       }
 
       return;
--- a/src/app/components/editor/output.ts
+++ b/src/app/components/editor/output.ts
@@ -212,15 +212,12 @@ export const getMentions = (mx: MatrixClient, roomId: string, editor: Editor): M
     if (node.type === BlockType.CodeBlock) return;
 
     if (node.type === BlockType.Mention) {
-      const mentionId = node.id;      
-      if (mentionId === getCanonicalAliasOrRoomId(mx, roomId)) {
-        if (!node.name.startsWith('#')) {
-          mentionData.room = true;
-        } 
-      } 
+      if (node.name === '@room') {
+        mentionData.room = true;
+      }
       
-      if (isUserId(mentionId) && mentionId !== mx.getUserId()) {
-        mentionData.users.add(mentionId);
+      if (isUserId(node.id) && node.id !== mx.getUserId()) {
+        mentionData.users.add(node.id);
       }
 
       return;


Santhoshkumar044:

src/app/components/editor/output.ts
```diff
@@ -212,11 +212,15 @@ export const getMentions = (mx: MatrixClient, roomId: string, editor: Editor): M
...
 
     if (node.type === BlockType.Mention) {
-      if (node.id === getCanonicalAliasOrRoomId(mx, roomId)) {
-        mentionData.room = true;
-      }
-      if (isUserId(node.id) && node.id !== mx.getUserId()) {
-        mentionData.users.add(node.id);
+      const mentionId = node.id;      
+      if (mentionId === getCanonicalAliasOrRoomId(mx, roomId)) {
+        if (!node.name.startsWith('#')) {
```
ajbura: i think it would be better to exect match it for `@room`

ajbura (changes requested):

src/app/components/editor/output.ts
Santhoshkumar044: Thanks for the suggestion I will update the logic to use an exact match for @room and push the changes shortly

Santhoshkumar044: Updated to use exact match on @room as suggested, tested and working correctly

Santhoshkumar044:
Hi @ajbura 
I have addressed the requested changes and updated it, The latest Netlify preview reflects the fixes.
Ready for re-review.Thanks!

kfiven (approved):

ajbura (approved):

PR #2587, opened 2026-02-03
fix: set m.fully_read marker when marking rooms as read

andrewmurphyio:
Previously `markAsRead()` only sent `m.read` receipts via `sendReadReceipt()`. This meant the read position was not persisted across page refreshes, especially noticeable in bridged rooms (Slack, Discord bridges).

Now uses `setRoomReadMarkers()` which sets both:
- `m.fully_read` marker (persistent read position)
- `m.read` receipt

## Problem
When viewing a room and then refreshing the page, the room would still show as unread. The `m.read` receipt was being sent (visible in network tab), but without `m.fully_read`, the client has no persistent record of where you read up to.

## Solution
Use `setRoomReadMarkers()` instead of `sendReadReceipt()`. This single call sets both markers, ensuring read state persists across sessions.

### DIFF
--- a/src/app/utils/notifications.ts
+++ b/src/app/utils/notifications.ts
@@ -1,6 +1,6 @@
 import { MatrixClient, ReceiptType } from 'matrix-js-sdk';
 
-export async function markAsRead(mx: MatrixClient, roomId: string, privateReceipt: boolean) {
+export async function markAsRead(mx: MatrixClient, roomId: string, privateReceipt?: boolean) {
   const room = mx.getRoom(roomId);
   if (!room) return;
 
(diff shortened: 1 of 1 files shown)


kfiven (approved):

ajbura (approved):

kfiven:
@andrewmurphyio @ajbura I have reverted this as notifications were not clearing after this change.

Wqrld:
I believe this should still be applied in some way. While debugging the same issue i came up with the following almost identical solution:
```diff
diff --git a/src/app/utils/notifications.ts b/src/app/utils/notifications.ts
index a23bd1a..8aa8726 100644
--- a/src/app/utils/notifications.ts
+++ b/src/app/utils/notifications.ts
@@ -1,4 +1,4 @@
-import { MatrixClient, ReceiptType } from 'matrix-js-sdk';
+import { MatrixClient } from 'matrix-js-sdk';
 
 export async function markAsRead(mx: MatrixClient, roomId: string, privateReceipt: boolean) {
   const room = mx.getRoom(roomId);
@@ -19,8 +19,15 @@ export async function markAsRead(mx: MatrixClient, roomId: string, privateReceip
   const latestEvent = getLatestValidEvent();
   if (latestEvent === null) return;
 
-  await mx.sendReadReceipt(
-    latestEvent,
-    privateReceipt ? ReceiptType.ReadPrivate : ReceiptType.Read
+  const latestEventId = latestEvent.getId();
+  if (!latestEventId) return;
+
+  // Set both the read receipt AND the fully_read marker
+  // The fully_read marker is what persists your read position across sessions
+  await mx.setRoomReadMarkers(
+    roomId,
+    latestEventId, // m.fully_read marker
+    privateReceipt ? undefined : latestEvent, // m.read (public, omitted when hiding activity)
+    latestEvent // m.read.private
   );
 }
```

I think you version withheld the private receipt and always sent the public one, and the private version was (i think) wrong.

I could not test this diff since all my phantom messages are unstuck already, but i believe this PR already had it 90% right.

PR #2588, opened 2026-02-04
fix: prevent phantom badges in bridged rooms

### LINKED ISSUE #2589: Phantom unread badges in bridged rooms
## Description
Bridged rooms (e.g., Slack, Discord, Telegram bridges via mautrix) show phantom unread badges even when there are no unread messages.

## Steps to Reproduce
1. Set up a bridged room (e.g., using mautrix-slack)
2. Receive messages in the bridged room
3. Read the messages on the bridge side (e.g., in Slack directly)
4. Open/refresh Cinny
5. The bridged room shows an unread badge despite having no unread messages

## Expected Behavior
No unread badge should appear if the server reports 0 unread notifications.

## Actual Behavior
Phantom badge appears because `roomHaveUnread()` returns `true` based on client-side read receipt comparison, even though the server notification count is 0.

## Root Cause
When reading messages on the bridge side, the read receipt doesn't sync back to Matrix. The client-side `roomHaveUnread()` function checks if there are messages after the local read marker, which returns `true` even when the server says there are no notifications.

## Environment
- Cinny version: dev
- Bridge: mautrix-slack (likely affects all bridges)
- Browser: Chrome

andrewmurphyio:
Fixes #2589

Fixes phantom unread badges appearing in bridged rooms (e.g., Slack, Discord bridges) where read receipts do not sync back to Matrix.

## Problem
When using bridged rooms, reading messages on the other side (e.g., in Slack) does not update the Matrix read receipt. This causes `roomHaveUnread()` to return `true` even when the server notification count is 0, resulting in phantom badges.

## Solution
Make the server notification count authoritative. Only add a room to the unread list if the server actually reports notifications (`total > 0` or `highlight > 0`). This prevents phantom badges while still respecting the server's notification state.

## Testing
1. Set up a bridged room (e.g., Slack bridge)
2. Read messages on the bridge side (not in Cinny)
3. Refresh Cinny
4. Previously: phantom badge would appear
5. Now: no badge if server reports 0 notifications

### DIFF
--- a/src/app/utils/room.ts
+++ b/src/app/utils/room.ts
@@ -248,7 +248,12 @@ export const getUnreadInfos = (mx: MatrixClient): UnreadInfo[] => {
     if (getNotificationType(mx, room.roomId) === NotificationType.Mute) return unread;
 
     if (roomHaveNotification(room) || roomHaveUnread(mx, room)) {
-      unread.push(getUnreadInfo(room));
+      const info = getUnreadInfo(room);
+      // Only show badge if server actually reports notifications
+      // This prevents phantom badges in bridged rooms where read receipts don't sync
+      if (info.total > 0 || info.highlight > 0) {
+        unread.push(info);
+      }
     }
 
     return unread;


vrisalab:
I've been testing this for a week, and it fixes the issue on a single client. Sadly, when using multiple devices (Desktop and a phone for example), ghost notifications for things read on one device appear on the other device. I'm not sure if this is easily fixable, though.

PR #2599, opened 2026-02-11
feat: Add video call support

### LINKED ISSUE #528: Video and Voice Calling
### Describe the problem

Currently, there is no way to voice call someone in Cinny.

An option to voice call ppl :)

### Alternatives considered

Using another matrix client

### Additional context

_No response_

hazre:
### Description
Fixes #528

Revives and continues #2335 (Element Call support), as the original PR has stalled.

Merged `dev` into this branch, resolved conflicts, and incorporated patches from @YoJames2019 to bring everything up to date and reviewable.

Marked as draft for now since some UX/polish improvements are still needed.

Credits to @GigiaJ for the original implementation and @YoJames2019 for follow up patches.

#### Need to fix (help would be appreciated):
- [ ] User doesn't show up when joining the call sometimes.
- [ ] Make E2EE Voice rooms work.
- [x] Voice Room Permissions 
- [ ] Some kind of UI for homeserver's that don't support call features.
- [ ] Connecting/disconnecting to voice rooms takes seconds.
- [x] Better new room creation experience, Have voice rooms as different type.
- [x] Connecting status.
- [x] Disable Webcam by default.
- [ ] Add our own Call controls instead of default Element call controls.
- [ ] Deafen functionality. (If possible, haven't investigated)

#### Nice to haves:
- [ ] Show a badge on space if in a voice room.
- [ ] User voice status in the room sidebar. (Blocked by official Element-call)
- [ ] User Muted Status in room sidebar. (Blocked by official Element-call)
- [ ] widget Folds/Cinny theming. (Blocked by official Element-call)
- [ ] Request Camera permissions only on enabling camera.
- [ ] Show duration of the call in the call status navigation. 

---

#### Type of change

### Checklist:

Preview: https://2599--pr-cinny.netlify.app/

Any feedback/Questions, please post in the dedicated thread: https://github.com/cinnyapp/cinny/pull/2599#discussion_r2802460116


hazre:

YoJames2019:
WOW that was fast, good job

romancenoire:
Insane work 🥳

hazre:
@YoJames2019 btw if u wanna talk about the PR, I'm on cinny's matrix space https://matrix.to/#/#cinny-space2:matrix.org

YoJames2019:
> @YoJames2019 btw if u wanna talk about the PR, I'm on cinny's matrix space https://matrix.to/#/#cinny-space2:matrix.org

MatrixError: [403] You are not invited to this room.

hazre:
> > @YoJames2019 btw if u wanna talk about the PR, I'm on cinny's matrix space https://matrix.to/#/#cinny-space2:matrix.org
> 
> MatrixError: [403] You are not invited to this room.

Oh my bad, I forgot they have a weird captcha thing. Try this maybe? https://matrix.to/#/#cinny:matrix.org

a55uka:
does this also implement the legacy calling in cinny?

hazre:
> does this also implement the legacy calling in cinny?

not sure what you mean by legacy calling. This implements https://github.com/element-hq/element-call

a55uka:
'legacy calling' is what the element calls just a TURN server being run (coturn etc), i have no idea if the element calling stack actually implements the 'legacy' calling, alot of servers ive come across dont actually seem to run the call stack but just a turn server

GimleLarpes:
AFAIK it only implements modern element call, not the old workaround of using external call services. It's been a while since I looked at this stuff though, so might be wrong.

Mondrethos:
@hazre @YoJames2019 thank you for everything!

lizard-demon:
BRAVO

package.json
```diff
@@ -54,6 +56,7 @@
     "jotai": "2.6.0",
     "linkify-react": "4.1.3",
     "linkifyjs": "4.1.3",
+    "matrix-widget-api": "1.11.0",
```
wrenix: why such an old version?

package.json
```diff
@@ -24,6 +25,7 @@
     "@atlaskit/pragmatic-drag-and-drop-auto-scroll": "1.3.0",
     "@atlaskit/pragmatic-drag-and-drop-hitbox": "1.0.3",
     "@fontsource/inter": "4.5.14",
+    "@matrix-org/react-sdk-module-api": "2.5.0",
```
wrenix: should we use an archived project?

Mondrethos:
> https://2599--pr-cinny.netlify.app

Works great so far.

GigiaJ:
Appreciate you picking this up. Apologies. I've been busy between work, illness, and a number of other project commitments.

lavenderdotpet:
> > https://2599--pr-cinny.netlify.app
> 
> Works great so far.

very exciting 
will be nice to have a really good clients to push discord users to once this is added

Mondrethos:
Yeah, working really great, putting it through the paces. I noticed that on my phone since I'm testing it on several devices, it doesn't start my camera automatically like element X does, so I really appreciate that.

xetem:
> Preview: https://2599--pr-cinny.netlify.app
> ⚠️ Exercise caution. Use test accounts. ⚠️

I noticed a bug, when opening a room that's a part of a space, like a bridged DM or a room that's part of two pinned spaces, opening it navigates to the outer most space.

To try and elaborate, if I open a DM that's bridged from Discord from the "Direct Messages" tab, it will navigate the sidebar to the Discord root space and open the room from there.
Likewise, if I open a room from a pinned space, again, Discord bridged guild is a good way to trigger this, opening a room navigates again to the root Discord space and opens the room from there instead of wherever in navigation you currently are.

lavenderdotpet:
> > Preview: https://2599--pr-cinny.netlify.app
> > ⚠️ Exercise caution. Use test accounts. ⚠️
> 
> I noticed a bug, when opening a room that's a part of a space, like a bridged DM or a room that's part of two pinned spaces, opening it navigates to the outer most space.
> 
> To try and elaborate, if I open a DM that's bridged from Discord from the "Direct Messages" tab, it will navigate the sidebar to the Discord root space and open the room from there.
> Likewise, if I open a room from a pinned space, again, Discord bridged guild is a good way to trigger this, opening a room navigates again to the root Discord space and opens the room from there instead of wherever in navigation you currently are.


does this not happen in main?

xetem:
> > > Preview: https://2599--pr-cinny.netlify.app
> > > ⚠️ Exercise caution. Use test accounts. ⚠️
> > 
> > 
> > I noticed a bug, when opening a room that's a part of a space, like a bridged DM or a room that's part of two pinned spaces, opening it navigates to the outer most space.
> > To try and elaborate, if I open a DM that's bridged from Discord from the "Direct Messages" tab, it will navigate the sidebar to the Discord root space and open the room from there.
> > Likewise, if I open a room from a pinned space, again, Discord bridged guild is a good way to trigger this, opening a room navigates again to the root Discord space and opens the room from there instead of wherever in navigation you currently are.
> 
> does this not happen in main?

No, it maintains the navigation wherever the room was selected for on stable v4.10.2 and I just tested on the nightly that was pushed with 83e5125.

Mondrethos:
Since there's a lot of chatter online over this PR due to the Discord situation, so that non technical people can understand better, what exactly are we missing before this PR is merged into Cinny and live for the masses?

hazre:
Added call connecting status:

https://github.com/user-attachments/assets/974d0e6e-4a99-4d2f-a2c4-e31b35f4187d

src/app/features/create-room/CreateRoom.tsx
```diff
@@ -170,6 +184,20 @@ export function CreateRoomForm({ defaultKind, space, onCreate }: CreateRoomFormP
...
       </Box>
+      <SequenceCard
+        style={{ padding: config.space.S300 }}
+        variant="SurfaceVariant"
+        direction="Column"
+        gap="500"
+      >
+        <SettingTile
+          title="Call Room"
+          description="Enable this to create a room optimized for voice calls."
```
1vel:
```suggestion
          description="Optimize this room for voice calls."
```

hazre: I'm thinking of redoing that UI in general, have a dedicated Selection of Room Types at the top. Instead of this toggle that can easily be missed.

hazre:
> Since there's a lot of chatter online over this PR due to the Discord situation, so that non technical people can understand better, what exactly are we missing before this PR is merged into Cinny and live for the masses?

It really depends, the bare bones functionality works using the element call widget (basically like element web client) but it's nowhere polished enough. 

Besides that, the main maintainers seem to be working on a new sdk and that most likely will take precedence over this. So lots of this PR would have to be rewriten more or less. And maybe even remove the widget solution for a custom implementation instead. But that's very much up in the air right now.

At least that's what we briefly discussed in the Cinny matrix space.

Mondrethos:
> > Since there's a lot of chatter online over this PR due to the Discord situation, so that non technical people can understand better, what exactly are we missing before this PR is merged into Cinny and live for the masses?
> 
> It really depends, the bare bones functionality works using the element call widget (basically like element web client) but it's nowhere polished enough.
> 
> Besides that, the main maintainers seem to be working on a new sdk and that most likely will take precedence over this. So lots of this PR would have to be rewriten more or less. And maybe even remove the widget solution for a custom implementation instead. But that's very much up in the air right now.
> 
> At least that's what we briefly discussed in the Cinny matrix space.

So this has no chance of being merged in in the next few weeks for all the people who are trying to move from Discord with Cinny as their client of choice?

vite.config.js
```diff
@@ -13,6 +13,10 @@ import buildConfig from './build.config';
 
 const copyFiles = {
   targets: [
+    {
```
kfiven, line 16: Feedback thread. Please leave any feedback/questions here if it's not code review.

shootie22:
> > Since there's a lot of chatter online over this PR due to the Discord situation, so that non technical people can understand better, what exactly are we missing before this PR is merged into Cinny and live for the masses?
> 
> 
> 
> It really depends, the bare bones functionality works using the element call widget (basically like element web client) but it's nowhere polished enough. 
> 
> 
> 
> Besides that, the main maintainers seem to be working on a new sdk and that most likely will take precedence over this. So lots of this PR would have to be rewriten more or less. And maybe even remove the widget solution for a custom implementation instead. But that's very much up in the air right now.
> 
> 
> 
> At least that's what we briefly discussed in the Cinny matrix space.

I would argue that pushing this through regardless of the new SDK and slight lack of polish should be done asap, because of the Discord situation. I would rather have an alternative that works and is less polished than no alternative.  

The timing here could really work in Cinny's and Matrix's advantage, providing a good experience for users moving over. Seeing no call functionality is much worse than some missing UI/UX polish.

vite.config.js
ghost: I have people on discord who I am waiting for cinny to merge this before joining. They highly value voice calls and cinny is the client I would like to them to use. Is the timeline too long and I should suggest other clients? Or is it possible just to wait?

Mondrethos: I'm also curious about this. The UI already looks incredible and the people I've shown it to are very happy with it. Anything else can be added in as updates.

xetem: Well there's the [bug I brought up](#issuecomment-3888870398), which as of 92f490e9d9ee9a34362e46713e808d41dfc6d612 is still occurring, and is a regression from this PR as it's not happening on dev or latest.

I also don't see any place in settings or the UI to specify your devices (mic/camera), being able to set those up ahead of time would be very beneficial for when our OS doesn't want to respect our specified device defaults.

There's also a lack of calling from non-call specific rooms. Element allows calls from any room, not just specifically designated "call rooms." Which with the current state of [MSC3417](https://github.com/matrix-org/matrix-spec-proposals/pull/3417) really muddies the water on what's intended. Having call UX parity with Element itself I personally think is should be a baseline, which would include initiating and accepting calls from any kind of room except spaces. And as of right now, if an Element user calls a Cinny user with Element Call, this PR currently does not ring at all, which would indicate to a user that the client doesn't support the feature.

henk717:
Will this allow for calling in DM's? That is the main thing I am missing that is keeping me stuck to Element.

vite.config.js
lizard-demon: Element allowing calling from any room is exactly what I dislike about element. It's confusing and clunky.

YoJames2019: I agree. I personally think calling from any room is a strange feature to have

YoJames2019:
> Will this allow for calling in DM's? That is the main thing I am missing that is keeping me stuck to Element.

not as of yet, possibly once we get element call polished up

vite.config.js
Mondrethos: Do we need it to follow Element's setup 1:1?

I know most people with entire communities would be happy with what we have now and then to receive more updates in the future, I wish we had communication from one of the lead maintainers like @ajbura

GigiaJ: Personally, I don't enjoy the thought of calling from any room. I didn't add it because frankly it isn't entirely practical. However, any room MUST be able to have a call (though I would find it acceptable to table the implementation for handling that for now).

It makes more sense, instead, to trust that the room will be aptly labeled as a room supporting calling (which I believe is the case even for Element) and we might wish to supply a much lighter indicator that a room has an active call going on (in a non-call room).

It has been about eight months since I added calling, so please feel free to disregard my thoughts there if they're not valid anymore.


On the notion of notifications for calling is an interesting one. It should be fairly easy to sort of placeholder an event notification for calling in non-supported rooms (so users are at least aware of the occurrence).

xetem: [MSC4196](https://github.com/matrix-org/matrix-spec-proposals/pull/4196) does define that calls are designed to support DMs, standard rooms, and dedicated rooms. I can someone understand people's hesitation with standard room calling, but DM calling is certainly essential for this feature to be usable, IMO. Having to basically room upgrade every DM to be of type `org.matrix.msc3417.call` just to make a call seems extremely arbitrary (especially because they'll need to be upgraded again at some point to be of type `m.call` _if_ that MSC is merged in).

To be frank, I really don't understand what the hesitation is. In the current version Matrix spec, there are two types of rooms: standard rooms and `m.space`. In any standard room, being able to send an `m.call` event is expected of clients that have the ability to make calls, and clients with the ability to receive calls should be able to understand an `m.call` event. Now, there are different kinds of `m.call` events, this PR isn't looking at all at legacy calls, nor should it, if Cinny were to support that it'd be through a different feature PR, and a different implementation (though could use the same UI).

The reason I bring up Element, and am seeking for feature parity with it, is because Matrix is supposed to be an inter-compatible ecosystem. I _hate_ that I have to make calls with Element, it's basically the only reason I still have it installed. And if this implementation doesn't do things the way that Element expects them to work, then it doesn't change my situation or the situation of anyone else in my position. I have friends that use Element, Cinny on mobile is honestly not that usable as a PWA, I don't know what it is about PWA's on Android but notifications basically are halted the second the app is closed. So I use FluffyChat there (along side Element because it too does not have working calls). Juggling different clients that have different features is honestly one of the main reasons why Matrix isn't as successful as it should be. I'm a power user, I know I'm a power user, and I can't in good faith recommend people join Matrix when I can't say "just download [some specific] client." The only client that I can recommend non-power users right now is Element, as much as _I_ don't like Element, because from a feature perspective the only thing it lacks is MSC2545 emotes.

Sorry for the long-winded reply, trust me, I cut a lot out. But I feel like I'm the odd one out here, I personally wouldn't call this PR "add[ing] Element Call" if you can't actually exchange in calls with Element users.

henk717: Yes interoperability is key. Switching people to matrix is hard enough, I am not going to then subject them to further restrictions so I will use what their app supports. 

If Cinny can't do DM calling in the end I simply wouldn't use it. If I am in a space where the habbit is to call in a specific room (Jitsi seems unsupported to even though its a widget) then I can't use cinny if it doesn't support that. So at minimum when these switch to element call it should support that.

I haven't used element call much yet since my friends don't have it. But from what I understood its supposed to auto create a dedicated call room when you call / accept?

If so exposing that button should be simple.

xetem: > I haven't used element call much yet since my friends don't have it. But from what I understood its supposed to auto create a dedicated call room when you call / accept?

No, it doesn't make a new room, it works very similarly to legacy 1-on-1 calls, sending an m.call event to an existing room, just using a different backend provider, LiveKit in this case.

Standalone Element Call does make new rooms for each call, but that's a different thing and runs on its own Synapse instance with its own client, not even Element X works with that kind of setup.

biscober: I think at minimum this implementation should support calls in DM in addition to the "call room" support that was added.  It's hard to move people over to the Matrix platform if they won't be able to do both 1:1 calls and have voice channels - that's the bare minimum for feature parity with other popular platforms.

Ideally, this implementation would support calling in _any_ non-space room (like Element) while continuing to support the "call rooms".  That would be the best case for interoperability.  I believe room admins can change permissions to prevent calling in "regular" rooms if they don't want that (which is the sane thing to do imo).

hazre:
I added a list of things that probably should be addressed before review to the PR body. If anyone wants to take up on any of them, it would be a great deal of help and would speed things up!

Sollace:
So I'm running this branch now and looking around I'm pretty happy with how it is. One thing, though, can we get the options to choose sound devices added here?

image

Being able to mute and unmute without being in the call is good, but I'd also like to be able to specify what sound device to use without joining a call too. Both for convenience and because trying to change sound devices whilst it's actively using one will tend to crash my browser (not blaming cinny or element - it's my browser that's at fault, just this would help to work around this issue).

TymekV:
I'd like to help with implementing voice room creation in the UI.

TymekV:
Do you think that it would be better to add another option in room creation dialog (text/voice) or to add a separate option for creating a voice room?

image

In my opinion this looks somewhat cluttered, I'm considering going with Element's approach (shown below)

image

LinuxinaBit:
> One thing, though, can we get the options to choose sound devices added here?

I think it's useful to look at what Discord does UI layout-wise here:
image
They use little dropdown menus next to the Mute and Deafen options to give access to quick audio settings (the options in the menu with the arrow open on mouse hover)
image
Since Element Call doesn't have a "Deafen" feature, this stuff could probably be integrated into a single audio quick menu in Cinny's case:
image

Depending on exactly how Cinny integrates with Element call, the option your share ones screen might be a good idea to add to the toolbar as well.

The big question is if this is all necessary, or is a shortcut button to Element Call's audio settings good enough?
Personally I don't really know, and that's a decision for all of you who are actually making this.

---
Also I love the progress here, it already looks great.
Good work to everyone involved!

hazre:
> So I'm running this branch now and looking around I'm pretty happy with how it is. One thing, though, can we get the options to choose sound devices added here?
> image
> 
> Being able to mute and unmute without being in the call is good, but I'd also like to be able to specify what sound device to use without joining a call too. Both for convenience and because trying to change sound devices whilst it's actively using one will tend to crash my browser (not blaming cinny or element - it's my browser that's at fault, just this would help to work around this issue).

I'm not sure we have control over that. 

> Do you think that it would be better to add another option in room creation dialog (text/voice) or to add a separate option for creating a voice room?
> image
> 
> In my opinion this looks somewhat cluttered, I'm considering going with Element's approach (shown below)
> image

Yes! That's what I meant with "Better new room creation experience, Have voice rooms as different type." item on the list. that's exactly how I imagined it. A context menu beforehand is fine with me as well.

vite.config.js
hazre: @xetem btw can you see if that issue you reported is fixed now? if not, could provide a screenshot/video of expected outcome and what you experiencing now? I don't have any bridged rooms to test with it, so I can't replicate it.

AndrewHack242: Does this PR include the ability to share screen / turn on camera?

YoJames2019: > Does this PR include the ability to share screen / turn on camera?

yes

xetem: @hazre It's fixed now, thanks!

hazre: > @hazre It's fixed now, thanks!

Glad to hear :)

CoderPenguin1-dev: Speaking of share screen, is it planned to actually select what screen you're actually viewing, similar to Commet?  
My main issue with Element and the current state of the PR is exactly that. as my community does use multiple shares at once. 
I'm somewhat assuming it will be, judging by the PR's description itself, but it'd still be nice to know for sure.
Also sorry if this is somehow the wrong place to ask.

TymekV: > Speaking of share screen, is it planned to actually select what screen you're actually viewing, similar to Commet? My main issue with Element and the current state of the PR is exactly that. as my community does use multiple shares at once. I'm somewhat assuming it will be, judging by the PR's description itself, but it'd still be nice to know for sure. Also sorry if this is somehow the wrong place to ask.

For now this PR uses Element Calls, so it will behave exactly like Element, but if we add a custom calling ui sometimes in the future it will likely change. You can also try opening an issue on the [Element Call](https://github.com/element-hq/element-call) repository.

CoderPenguin1-dev: > For now this PR uses Element Calls, so it will behave exactly like Element, but if we add a custom calling ui sometimes in the future it will likely change. You can also try opening an issue on the [Element Call](https://github.com/element-hq/element-call) repository.

I'll look into doing that. Thanks for the response.

LoEhyeh: Looking forward to this!

Sollace:
> Do you think that it would be better to add another option in room creation dialog (text/voice) or to add a separate option for creating a voice room?
> image
> 
> In my opinion this looks somewhat cluttered, I'm considering going with Element's approach (shown below)
> image

A minor nitpick: This dialog seems to take up a lot of space. I'm looking at it on my screen and it's forcing the page to scroll, and I feel like the order could be more efficient.

Any thoughts on doing something like this instead?

image

With the name field being the default focus it makes sense to put it as the first input. You want to minimise people having to scan back and forth with their eyes when reading the form. The existing order forces you to start in the middle, scan up to change the type, then scan down to change the access, and other options.

Name |-- Access --> Type -| -> Access -- Name -> Topic -> Options -> Create

Natural reading order should be top to down.

Name -> Topic -> Type -> Access -> Create

Edit: Other things like Federation and Knock to Join could be rolled into the access selection, but that may is probably out of scope of this PR.

NextWork123:
> > Do you think that it would be better to add another option in room creation dialog (text/voice) or to add a separate option for creating a voice room?
> > image
> > In my opinion this looks somewhat cluttered, I'm considering going with Element's approach (shown below)
> > image
> 
> A minor nitpick: This dialog seems to take up a lot of space. I'm looking at it on my screen and it's forcing the page to scroll, and I feel like the order could be more efficient.
> 
> Any thoughts on doing something like this instead?
> 
> image
> With the name field being the default focus it makes sense to put it as the first input. You want to minimise people having to scan back and forth with their eyes when reading the form. The existing order forces you to start in the middle, scan up to change the type, then scan down to change the access, and other options.
> 
> Name |-- Access --> Type -| -> Access -- Name -> Topic -> Options -> Create
> 
> Natural reading order should be top to down.
> 
> Name -> Topic -> Type -> Access -> Create
> 
> Edit: Other things like Federation and Knock to Join could be rolled into the access selection, but that may is probably out of scope of this PR.

not bad at all i would like small icons on the right picture. It seem a bit distracting at me

Sollace:
> not bad at all i would like small icons on the right picture. It seem a bit distracting at me

I made the icons larger to match the height of the two lines but I tentatively agree, they might be a little too large. The image is just a mockup to demonstrate the concept so whoever picks it up can use their discretion as to how they implement the specific details.

TymekV:
And what about this modal when creating a room in a space? Three oprions horizontally would look weird so I'd suggest to just reorder them to match this new design. And I would implement the concept, just make icons a little bit smaller. What do you think about this?

Sollace:
> And what about this modal when creating a room in a space? Three oprions horizontally would look weird so I'd suggest to just reorder them to match this new design. And I would implement the concept, just make icons a little bit smaller.

What I imagine is like a horizontal switch where the highlight moves/slides to the option you click. Maybe to save space you only show the text for the option that's currently select? So the other two only take up the size of the icon.

> What do you think about this?

Go for it

vite.config.js
Mondrethos: Looking great so far!

Is the Dev link being updated with the latest merges to this PR?

Mondrethos:
Is there a new automated dev link with all the latest changes for preview and testing/feedback? I can't seem to find it.

hazre:
> Is there a new automated dev link with all the latest changes for preview and testing/feedback? I can't seem to find it.

Under the PR body you can find it, but it's this: https://2599--pr-cinny.netlify.app/

CEbbinghaus:
> Under the PR body you can find it, but it's this: https://2599--pr-cinny.netlify.app/

I just tested the latest version and it does not seem to have call support for arbitrary text channels / DMs. This to me is the #​1 most important feature of adding Voice/Video call support. 

I do love the way it shows users active in a voice channel underneath, Although I think they could use a litte extra indentation to make sure they look like they are **part** of the channel rather than their own entry in the list of chats.

image

vite.config.js
Mondrethos: @hazre For the first post is E2EE voice rooms complete? In the preview link I'm joining. encrypted voice rooms.

Nova1545:
Based on the live dev link it seems to be nice, very discord-esq room creation for text/voice channels

src/app/features/room/RoomViewHeader.tsx
```diff
@@ -381,69 +382,75 @@ export function RoomViewHeader() {
...
+                        {pinnedEvents.length}
+                      </Text>
+                    </Badge>
+                  )}
+                  <Icon size="400" src={Icons.Pin} filled={!!pinMenuAnchor} />
+                </IconButton>
+              )}
+            </TooltipProvider>
+          )}
+          {(!room.isCallRoom() || isChatOpen) && (
```
CEbbinghaus: This Condition is a duplicate of the one above, The items should probably just be Wrapped in a fragment to ensure the conditions don't go out of sync.

src/app/features/room/RoomViewHeader.tsx
```diff
@@ -381,69 +382,75 @@ export function RoomViewHeader() {
...
-                  isKeyBackward: (evt: KeyboardEvent) => evt.key === 'ArrowUp',
-                  escapeDeactivates: stopPropagation,
-                }}
-              >
-                <RoomPinMenu room={room} requestClose={() => setPinMenuAnchor(undefined)} />
-              </FocusTrap>
-            }
-          />
-          {screenSize === ScreenSize.Desktop && (
+          {(!room.isCallRoom() || isChatOpen) && (
```
CEbbinghaus, line 374: A Call room can contain messages, So is it unable to contain Pinned messages? I thought those were just stored in the room state?

src/app/features/room/RoomViewHeader.tsx
```diff
@@ -454,12 +461,35 @@ export function RoomViewHeader() {
...
+                </Tooltip>
+              }
+            >
+              {(triggerRef) => (
+                <IconButton fill="None" ref={triggerRef} onClick={toggleChat}>
+                  <Icon size="400" src={Icons.Message} filled={isChatOpen} />
+                </IconButton>
+              )}
+            </TooltipProvider>
+          )}
```
CEbbinghaus, line 491: Can we add the opposite to this as well? So for Non-Call rooms we have a call icon? This is what Element does and Discord supports. So **especially** for all the people migrating from Discord, it seems prudent to follow convention.

src/app/features/room/Room.tsx
```diff
@@ -40,8 +42,17 @@ export function Room() {
...
+            {room.isCallRoom() && screenSize === ScreenSize.Desktop && (
+              <Line variant="Background" direction="Vertical" size="300" />
+            )}
+            <RoomView room={room} eventId={eventId} />
+          </Box>
+        </Box>
+        {!room.isCallRoom() && screenSize === ScreenSize.Desktop && isDrawer && (
           <>
             <Line variant="Background" direction="Vertical" size="300" />
             <MembersDrawer key={room.roomId} room={room} members={members} />
```
CEbbinghaus, line 60: Do we not want a Members Drawer for Call rooms?

CEbbinghaus (changes requested):
Just a few UX comments

src/app/features/room/RoomViewHeader.tsx
xetem: The main difference I've noticed between call rooms and standard rooms, as the MSC is interpreted by Element, is that a call room always has a call available for users to join, instead of needing to make an ad-hoc call like in normal rooms. They still have normal message logs, and presumably can contain pinned messages as well. Though I haven't tested that myself.

src/app/features/room/Room.tsx
ajbura: they can also be viewable from room settings

xetem: It would make sense to, possibly collapsed by default, taking the same drawer space as the chat log.

src/app/features/room/RoomViewHeader.tsx
KittyKatt: From my brief testing, yes they are capable of having pinned messages and every other capability of a normal text room.

src/app/features/room/RoomViewHeader.tsx
Sollace: > This is what element does and *Discord Supports*

Can you clarify what you mean by this? Last I checked I didn't think Discord had call buttons anywhere other than DMs and dedicated voice channels.

Unless do you mean adding a icon to user names to indicate that they're in a call (which there's a plugin to do that on Discord)

src/app/features/room/RoomViewHeader.tsx
CEbbinghaus: Then we should probably not gate any of these features arbitrarily

src/app/features/room/RoomViewHeader.tsx
CEbbinghaus: Both discord and element have a call button in; DMS, rooms/groupchats.

The only place they differ is in Spaces/Servers.
Discord allows calls in only voice channels within servers. Whereas matrix just treats everything as rooms with some rooms part of a space.

This means that in element ALL rooms have call support, which includss every text channel in a Space, with Discord its only dedicated voice channels.

In any case. DMs and Rooms (not part of spaces) should have a call button in 100% of cases. While I would argue that within spaces we should follow elements lead of just turning on voice for all.

One thing to consider is that calling can have a required power level, and we should grey out the call button if the user doesn't meet this requirement.

src/app/features/room/Room.tsx
CEbbinghaus: Collapse by default is a very sane approach IMO

src/app/features/room/RoomViewHeader.tsx
YoJames2019: it isnt gated arbitrarily, the pinned messages button is not shown unless the text chat tab is open. you can see pinned messages in a call room if you have the chat tab open

CEbbinghaus: Oh? I hadn't noticed. Let me double check but that is very sane if the case

src/app/components/create-room/CreateRoomVoiceSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { SettingTile } from '../setting-tile';
+import { CreateRoomVoice } from './types';
+
+type CreateRoomVoiceSelectorProps = {
+  value?: CreateRoomVoice;
+  onSelect: (value: CreateRoomVoice) => void;
+  disabled?: boolean;
+  getIcon: (kind: CreateRoomVoice) => IconSrc;
+};
+export function CreateRoomVoiceSelector({
```
ajbura: i think we should rather call it `CreateRoomType**`

src/app/components/create-room/utils.ts
```diff
@@ -82,6 +83,44 @@ export const createRoomEncryptionState = () => ({
   },
 });
 
+export const createRoomCallState = () => ({
+  type: 'org.matrix.msc3401.call',
+  state_key: '',
+  content: {},
+});
+
+export const createPowerLevelContentOverrides = (
```
ajbura: i think we shouldn't be creating this utility and instead create something like `createCallRoomPowerLevelsState` because initial state will always remain constant for each room type. so we do not need overrides utils.

src/app/components/create-room/utils.ts
```diff
@@ -94,6 +133,7 @@ export type CreateRoomData = {
   knock: boolean;
   allowFederation: boolean;
   additionalCreators?: string[];
+  powerLevelContentOverrides?: IPowerLevels;
```
ajbura: we do not need to pass this here as we already have the room type based on which we can call our util `createVoiceRoomPowerLevelsState`

src/app/components/create-room/utils.ts
```diff
@@ -136,5 +180,15 @@ export const createRoom = async (mx: MatrixClient, data: CreateRoomData): Promis
...
 
+  if (data.powerLevelContentOverrides) {
+    const roomPowers = await mx.getStateEvent(result.room_id, StateEvent.RoomPowerLevels, '');
+    const updatedPowers = createPowerLevelContentOverrides(
+      roomPowers,
+      data.powerLevelContentOverrides
+    );
+
+    await mx.sendStateEvent(result.room_id, StateEvent.RoomPowerLevels as any, updatedPowers, '');
+  }
```
ajbura: we should be pushing the power level state to the `initialState` before creating the room.

src/app/utils/room.ts
```diff
@@ -257,24 +257,60 @@ export const getUnreadInfos = (mx: MatrixClient): UnreadInfo[] => {
...
-    return space ? icons.SpaceLock : icons.HashLock;
-  }
-  if (joinRule === JoinRule.Invite) {
-    return space ? icons.SpaceLock : icons.HashLock;
-  }
-  if (joinRule === JoinRule.Public) {
-    return space ? icons.SpaceGlobe : icons.HashGlobe;
+  roomType?: string,
+  joinRule?: JoinRule,
+  locked?: boolean
```
ajbura: new changes really complicated this function logic.

```ts
if (roomType === RoomType.Space) {
  // TODO: return icon by join rule
}
if (roomType === RoomType.Call) {
  // TODO: return icon by join rule
}
// TODO: return icon by join rule
```

also unsure what locked param is for?

src/app/features/create-room/CreateRoom.tsx
```diff
@@ -37,20 +37,40 @@ import {
...
+};
+
+const getCreateRoomVoiceToIcon = (kind: CreateRoomVoice) => {
+  if (kind === CreateRoomVoice.VoiceRoom) return Icons.VolumeHigh;
+  return Icons.Hash;
 };
 
 type CreateRoomFormProps = {
   defaultKind?: CreateRoomKind;
+  defaultVoice?: CreateRoomVoice;
```
ajbura: can rename it to `defaultType`

src/app/features/lobby/LobbyHeader.tsx
```diff
@@ -165,7 +165,7 @@ export function LobbyHeader({ showProfile, powerLevels }: LobbyHeaderProps) {
             <Box shrink="No">
               <BackRouteHandler>
                 {(onBack) => (
-                  <IconButton onClick={onBack}>
+                  <IconButton fill="None" onClick={onBack}>
```
ajbura: not sure why this change was made?

src/app/features/room-nav/RoomNavItem.tsx
```diff
@@ -208,6 +217,7 @@ const RoomNavItemMenu = forwardRef<HTMLDivElement, RoomNavItemMenuProps>(
     );
   }
 );
+RoomNavItemMenu.displayName = 'RoomNavItemMenu';
```
ajbura: not sure what it is doing?

src/app/features/room-nav/RoomNavUser.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const openProfile = useOpenUserRoomProfile();
+  const space = useSpaceOptionally();
+  const { isActiveCallReady, activeCallRoomId } = useCallState();
+  const isActiveCall = isActiveCallReady && activeCallRoomId === room.roomId;
+  const userId = callMembership.sender ?? '';
+  const avatarMxcUrl = getMemberAvatarMxc(room, userId);
+  const avatarUrl = avatarMxcUrl
+    ? mx.mxcUrlToHttp(avatarMxcUrl, 32, 32, 'crop', undefined, false, useAuthentication)
+    : undefined;
+  const getName = getMemberDisplayName(room, userId) ?? getMxIdLocalPart(userId);
```
ajbura: it think it should be `name`?

src/app/features/room-settings/permissions/usePermissionItems.ts
```diff
@@ -196,6 +209,7 @@ export const usePermissionGroups = (): PermissionGroup[] => {
 
     return [
       messagesGroup,
+      callSettingsGroup,
```
ajbura: this setting should not appear in non-call rooms?

src/app/features/room/Room.tsx
```diff
@@ -23,7 +25,7 @@ export function Room() {
   const [hideActivity] = useSetting(settingsAtom, 'hideActivity');
   const screenSize = useScreenSizeContext();
   const powerLevels = usePowerLevels(room);
-  const members = useRoomMembers(mx, room.roomId);
+  const members = useRoomMembers(mx, room?.roomId);
```
ajbura: i think this change is not needed?

src/app/features/room/RoomView.tsx
```diff
@@ -92,51 +93,59 @@ export function RoomView({ room, eventId }: { room: Room; eventId?: string }) {
...
-                  <Text align="Center">You do not have permission to post in this room</Text>
-                </RoomInputPlaceholder>
-              )}
-            </>
-          )}
-        </div>
-        {hideActivity ? <RoomViewFollowingPlaceholder /> : <RoomViewFollowing room={room} />}
-      </Box>
-    </Page>
+    (!room.isCallRoom() || isChatOpen) && (
```
ajbura: i think we should decide whether we render `RoomView` in `Room.tsx` rather than in here.

src/app/hooks/useStateEvents.ts
```diff
@@ -1,28 +1,35 @@
...
-import { Room } from 'matrix-js-sdk';
+import { MatrixEvent, Room } from 'matrix-js-sdk';
 import { StateEvent } from '../../types/matrix/room';
+import { useMatrixClient } from './useMatrixClient';
 import { useForceUpdate } from './useForceUpdate';
 import { useStateEventCallback } from './useStateEventCallback';
-import { getStateEvents } from '../utils/room';
 
-export const useStateEvents = (room: Room, eventType: StateEvent) => {
+export const useStateEvents = (rooms: Room[], eventType: StateEvent): number => {
```
ajbura: not sure why this change was made?

src/app/pages/client/space/Space.tsx
```diff
@@ -296,7 +298,7 @@ function SpaceHeader() {
                 escapeDeactivates: stopPropagation,
               }}
             >
-              <SpaceMenu room={space} requestClose={() => setMenuAnchor(undefined)} />
+              {space && <SpaceMenu room={space} requestClose={() => setMenuAnchor(undefined)} />}
```
ajbura: since space can not be undefined not sure why this is conditionalize?

ajbura (changes requested):

src/app/features/room/Room.tsx
ajbura: i think not displaying it make sense because people in active call can misinterpret it as members in call instead of call room?

src/app/utils/room.ts
Sollace: locked = cannot change?

src/app/components/create-room/utils.ts
ajbura: sorry we should rather pass it to power_level_content_override in /createRoom?

src/app/features/room/RoomViewHeader.tsx
CEbbinghaus: Confirmed this is the case. This feedback is no longer applicable

src/app/features/room/Room.tsx
CEbbinghaus: I disagree as it makes it entirely impossible to see which people are part of a room. Which means posting pictures / messages in text channel may be sent to people you wouldn't be aware of.

We could move the member list into the text channel side-menu same as the pinned messages list.

src/app/utils/room.ts
CEbbinghaus: Locked is strange, May be the reason why my call room renders strangely:
image

src/app/features/room/Room.tsx
Sollace: > I disagree as it makes it entirely impossible to see which people are part of a room. Which means posting pictures / messages in text channel may be sent to people you wouldn't be aware of.
> 
> We could move the member list into the text channel side-menu same as the pinned messages list.

Having the member list button only accessible from the chat view (i.e. when it is opened let it have its own header bar or section of the header bar separate from the main one would help emphasise that distinction.

src/app/utils/room.ts
hazre: It's a placeholder icon, until we get the new folds icons. unless you mean something else.

src/app/features/room/RoomViewHeader.tsx
GimleLarpes: I would say that the call button should just be hidden if the action is forbidden. This is the pattern used in Cinny. And the most common use case I can see is that Admins of a space only want calling in certain rooms, so showing an action they can't take like calling in a text-only room would be confusing to the user.

src/app/features/room/Room.tsx
CEbbinghaus: It already has its own header bar which contains the Pinned messages button. So adding the Member list into that should be fine. I'd be worried about the horizontal space. and **ESPECIALLY** on mobile. But It seems like an alright compromise. 

To be honest though, I still believe the member list should be in the main header bar rather than in the chat. We can easily see who is in the call in both the list of users under the channel entry, as well as the active call that would be **fullscreen** in your call channel view. So I don't think it would cause confusion with people thinking they were actively on the call. Both Teams, Discord *and* element let you view the list of users in the main call window. So I would **strongly** advise to follow established convention:
image

src/app/features/room/RoomViewHeader.tsx
CEbbinghaus: Although, I would add the Threads button into the header bar when the chat sidebar is opened as well. Element currently has two buttons (one for chat and one for threads) which IMO is not the greatest user experience, Since threads **are** part of chat, it should be hidden unless chat is active.

CEbbinghaus (commented):
Additionally I would like to raise that Element by default pops up a "Join Call" window showing you your camera (if enabled) and giving you the option to join the call. 

Image

This is a Cruical part of UX this PR is missing since Just clicking on the channel is enough to automatically join it, So trying to look at messages requires you first join voice which is just plain not possible sometimes. And if camera is enabled by default, Just clicking on the channel would be enough to join you **with an active camera**, Without you having meant to. This means a Missclick could cause a very embarrasing moment.

Here we need to decide more than anything whether we want to behave like Discord:
> clicking on the channel auto joins, but with camera **always** disabled. Text channel can be accessed **without** joining the voice first through an additional UI element:
> Image

or Element:
> Clicking on the channel offers option to join the channel, Or open text chat without joining.
> Since we auto preview what the user would be showing (Video if Video is enabled), the user knows exactly what will be shown to other users.

Personally, I lean towards the former for voice rooms **within** spaces, as they should behave more like Discord Servers

and towards the latter for Rooms outside of spaces. Having a call room not connected to a space is a strange occurrence to begin with.

edit:

Another gripe is that call rooms **outside** of spaces have the same list of users in the call shown under the channel as those **within** spaces. This means you list of groupchats can be filled with 5 random people since they just happen to have joined a call.

Again here I would like to distinguish between Space rooms and non-Space rooms

amazingca:
> Any thoughts on doing something like this instead?
> 
> image

I'm a little partial towards the one on the left since the UI is a little more fleshed out. I'm sure others are also more familiar with the channel creation layout on the left.

undu:
> > Any thoughts on doing something like this instead?
> > <snip>
> I'm a little partial towards the one on the left since the UI is a little more fleshed out. I'm sure others are also more familiar with the channel creation layout on the left.

I think that the left one will be easier for internationalization, as it leaves more space for the text in the buttons

hazre:
> This is a Cruical part of UX this PR is missing since Just clicking on the channel is enough to automatically join it, So trying to look at messages requires you first join voice which is just plain not possible sometimes. And if camera is enabled by default, Just clicking on the channel would be enough to join you **with an active camera**, Without you having meant to. This means a Missclick could cause a very embarrasing moment.

You have cam and mute controls in the bottom sidebar, Camera is not enabled by default, no prescreening was an intentional choice not be like conferencing (like zoom) and be more like discord (in the current state of the PR). But being able to view the side chat without joining the voice room is completely valid. Either way all these are placeholder, @ajbura has new designs that will completely change how UI for this feature looks like. (discussed in the cinny matrix space)

Also I just wanted to mention that our goal is not to copy element or other matrix clients. when we can improve the user experience and do better. That's the whole point of being a different client.

Mondrethos:
> > This is a Cruical part of UX this PR is missing since Just clicking on the channel is enough to automatically join it, So trying to look at messages requires you first join voice which is just plain not possible sometimes. And if camera is enabled by default, Just clicking on the channel would be enough to join you **with an active camera**, Without you having meant to. This means a Missclick could cause a very embarrasing moment.
> 
> You have cam and mute controls in the bottom sidebar, Camera is not enabled by default, no prescreening was an intentional choice not be like conferencing (like zoom) and be more like discord (in the current state of the PR). But being able to view the side chat without joining the voice room is completely valid. Either way all these are placeholder, @ajbura has new designs that will completely change how UI for this feature looks like. (discussed in the cinny matrix space)
> 
> Also I just wanted to mention that our goal is not to copy element or other matrix clients. when we can improve the user experience and do better. That's the whole point of being a different client.

Is the goal to wait for @ajbura's new designs before this is finalzied and merged in?

ajbura:
> Is the goal to wait for @ajbura's new designs before this is finalzied and merged in?

nope. this will merge as soon as current requested changes are resolved.

Mondrethos:
> > Is the goal to wait for @ajbura's new designs before this is finalzied and merged in?
> 
> nope. this will merge as soon as current requested changes are resolved.

Thanks for answering :)

RoflMyPancakes:
I've been following this and able to use this just fine when built for web over the last few days, but when I build cinny-desktop with cinny on this branch, joining voice rooms hangs indefinitely from cinny-desktop (shows "Joining"). 

I just thought saying this sooner might get more eyes on the cinny-desktop angle.

Works fine with chrome, firefox. 
But with cinny-desktop with Tauri it's behaving differently.

/public/element-call/index.html requests end up returning cinny's index instead.

Query params break the file lookup in cinny-desktop. I was able to test:
- /public/element-call/index.html => Resolves to Element Call 
- /public/element-call/index.html?embed=true => Resolves to Cinny (not found, fallback to index) 

I'm assuming (based on some research of tauri's localhost plugin) that Tauri's localhost resolver doesn't strip query parameters before looking up the file. It tries to find index.html?embed=true literally (including query params), fails, and falls back to the root index.html.

I've been able to fix it a couple different ways in my testing but that's the core problem I hit. I don't know what the most elegant solution is if I'm right about this being a problem. 

I was able to fix it in just cinny by changing to url.hash from url.search, which caused other problems but I could at least join voice rooms. But changing cinny-desktop to not use the tauri localhost plugin could be another path.

Maybe someone more experienced with cinny-desktop should run this branch through a test.

Edit: The solution for me was to patch Tauri's localhost plugin (though it seems 2.x of tauri includes a localhost plugin with this fix). This made the fix entirely on the cinny-desktop side and I can use this branch unchanged. 

  // Original plugin:
  asset_resolver.get(req.url().into())

  // Patched plugin:
  let path = req.url().split('?').next().unwrap_or(req.url()).into();
  asset_resolver.get(path)

With that change, cinny-desktop works with this branch. Without it, joining voice rooms fails (the element-call iframe fails)

src/app/components/create-room/CreateRoomVoiceSelector.tsx
TymekV: Wouldn't it be confused with `CreateRoomKind`? Shouldn't we also rename it?

src/app/features/room/RoomViewHeader.tsx
hazre: Threads aren't fully supported yet in cinny, so it's not relevant to this PR.

package.json
hazre: Updated

package.json
hazre: Removed it.

src/app/features/room/Room.tsx
hazre: for now, for consistency sake, I pushed a commit that makes the toggle available on desktop. we can visit this later after this initial pr is merged and new design comes.

src/app/components/create-room/utils.ts
YoJames2019: this is incase in the future we add functionality to alter permissions BEFORE room creation

src/app/utils/room.ts
YoJames2019: the locked param is for overriding the icon in the case of missing permissions. like in the case of calls, if the user does not have the join call permission, you just set that param when fetching the icon and it will show the correct icon

src/app/features/room-settings/permissions/usePermissionItems.ts
YoJames2019: well technically any room can have a call in it according to matrix

vite.config.js
RoflMyPancakes: I pulled in the latest changes from the branch but seems like there's regressions. I can't join rooms. It shows me in the room then it leaves after the 15 second timeout, never succeeding. I can click multiple rooms and show up in all of them, leaving after the 15 second time out. Haven't had time to look into why, just that it seems like there's breaking changes in the latest series of commits.

jeanbono:
The commit 30303fe4c236559dbbf107c5fd25c7e8d32d7205 break the calls for me.
I'm getting disconnected almost immediatly after joining.
Everything is working fine with the previous commit (881da5cb5d4a879643f2647adc4d6dafa99ab618)

hazre:
> The commit [30303fe](https://github.com/cinnyapp/cinny/commit/30303fe4c236559dbbf107c5fd25c7e8d32d7205) break the calls for me. I'm getting disconnected almost immediatly after joining. Everything is working fine with the previous commit ([881da5c](https://github.com/cinnyapp/cinny/commit/881da5cb5d4a879643f2647adc4d6dafa99ab618))

Yeah it looks like the element call version we are using, uses 1.13.0 version of widget api package. For now I will downgrade it.

hazre:
@jeanbono I've pushed some changes to address that, see if it works now

jeanbono:
> @jeanbono I've pushed some changes to address that, see if it works now

Looks good now :+1:

vrisalab:
I've been testing this, and I've found a bug: When you join a room, and then close the tab, your user never leaves the call, not even after a server restart. If the user was sharing their screen, the screen remains perpetually shared. I'm not sure if the behavior is the same on Element.

GigiaJ:
> shared. I'm not sure if the behavior is the same on Element.

You can do this on Element too.  Quirk of Element Call I imagine. User has to timeout from the LiveKit SFU. That is why a server restart doesn't DC them even unless you restart that too I believe.

Sollace:
> I've been testing this, and I've found a bug: When you join a room, and then close the tab, your user never leaves the call, not even after a server restart. If the user was sharing their screen, the screen remains perpetually shared. I'm not sure if the behavior is the same on Element.

Ah yeah, I've seen this as well in fact when my browser has crashed whilst in a call.

@GigiaJ Is it possible to add something in to trigger the user leaving the call properly when the tab is closed?

i.e. Using the [beforeunload window event](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) or similar?

chxreyes:
Is the preview up to date with the current PR diff? Calls do not seem to work for me using the preview.

xetem:
> Is the preview up to date with the current PR diff? Calls do not seem to work for me using the preview.

Yeah I'm getting stuck trying to join a call in a call room without it ever resolving, can join the room successfully on Element.

kfiven:
Merging this in voice-calls PR for further testing. https://github.com/cinnyapp/cinny/pull/2680

PR #2309, opened 2025-04-19
Show image viewer when clicking url preview thumbnail

LeaPhant:
### Description
Show an image viewer when clicking the thumbnail in the url preview, similar to Element. Uses the same code as image content, maybe some deduplication necessary.

showcase of image preview

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #2633, opened 2026-02-15
Open full image in a new tab when clicking on url preview thumbnail

piko-piko:
Middle click on the url preview thumbnail opens the full image in a new window.
Under the hood it fetches the image from the proxy server, since to access it we need the client's auth token.


piko-piko:
I've made an attempt to rebase the branch, but CLA is still not pleased. Not sure what to do. Otherwise i would love a review - i'm not a React dev so i'm not sure what are the best practices for placing helper functions and i can work with a feedback

kfiven:
Just look at clabot comment. It has instructions.

piko-piko:

piko-piko:
@kfiven Thank you. I've got tunnel vision on the issue with the wrong authorship that i didn't properly internalized the instructions

src/app/components/url-preview/UrlPreview.tsx
```diff
@@ -3,13 +3,54 @@ import classNames from 'classnames';
...
+import { useMatrixClient } from '../../hooks/useMatrixClient';
+
+// mxcToHttp at ../../utils/matrix implements deprecated endpoint
+// GET /_matrix/media/v3/download/{serverName}/{mediaId}
+function mxcToHttp(mxcUrl: string, baseUrl: string) {
+  const match = mxcUrl.match(/^mxc:\/\/([^/]+)\/(.+)$/);
+  if (!match) return null;
+  const [, server, mediaId] = match;
+  return `${baseUrl}/_matrix/client/v1/media/download/${server}/${mediaId}`;
+}
```
ajbura: you can use `mxcUrlToHttp` utility for this which already exists in `src/app/utils/matrix.ts`

src/app/components/url-preview/UrlPreviewCard.tsx
```diff
@@ -34,7 +34,7 @@ export const UrlPreviewCard = as<'div', { url: string; ts: number }>(
 
       return (
         <>
-          {imgUrl && <UrlPreviewImg src={imgUrl} alt={prev['og:title']} title={prev['og:title']} />}
+          {imgUrl && <UrlPreviewImg mxcUrl={prev['og:image']} src={imgUrl} alt={prev['og:title']} title={prev['og:title']} />}
```
ajbura, line 45: rendering `ImageViewer` on an Modal Overlay will be a good idea instead of opening it into the new window. you can look at the `src/app/components/user-profile/UserHero.tsx` to see how it is done in user profile.

ajbura (changes requested):

src/app/components/url-preview/UrlPreviewCard.tsx
piko-piko: Thank you for the feedback! But opening it in a new windows is the whole point of the PR. I, and probably most of the powerusers, are used to how the browser handles displaying images natively - because it's a consistent way to look at images across sites, unlike modals on various sites which implement zooming or scrolling in various was (if at all). I'm just "porting" a feature, that i was able to do on discord - middle click on image to open it in a new tab. It's extra cumbersome on matrix, because image doesn't exist under an url, you have to use token to retrieve it from the proxy. 

On top of that there is already a PR, which implements modals for thumbnails so it would be duplicating someone's work https://github.com/cinnyapp/cinny/pull/2309

ajbura:
closing in favor of #2309

Issue #2353, opened 2025-06-03
Add "Copy Message" to message context menu


GimleLarpes:
### Describe the problem

Copying a message is a very common action. Therefore I suggest making this easier by adding it as a button in the message context menu. This button would copy the contents of the message (i.e. image data, text, etc.) to the clipboard.

### Describe the solution you'd like

A "Copy Message" button is added to the message context menu.
This button would copy the contents of the message to the clipboard.

GigiaJ:
I actually need this for mobile since highlighting messages on mobile is impractical to begin with (and removes the option for clean interaction with long-press)
I can clean up my PR for mobile context menus in a bit so there isn't any collision there (I refactor it a significant amount).

PR #2639, opened 2026-02-16
Add copy message to message context menu

TheDeveloperRyan:
### Description
This pull request adds a "Copy Message" button to the message context menu. When clicked, the button copies the body of the selected message to the users clipboard.

This feature addition was discussed in #2353  

Fixes

#### Type of change

### Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas (none needed)
- [x] I have made corresponding changes to the documentation (none needed)

### DIFF
--- a/src/app/features/room/message/Message.tsx
+++ b/src/app/features/room/message/Message.tsx
@@ -35,7 +35,7 @@ import { useHover, useFocusWithin } from 'react-aria';
 import { MatrixEvent, Room } from 'matrix-js-sdk';
 import { Relations } from 'matrix-js-sdk/lib/models/relations';
 import classNames from 'classnames';
-import { RoomPinnedEventsEventContent } from 'matrix-js-sdk/lib/types';
+import { type RoomPinnedEventsEventContent } from 'matrix-js-sdk/lib/types';
 import {
   AvatarBase,
   BubbleLayout,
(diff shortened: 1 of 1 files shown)


TheDeveloperRyan:

mmmykhailo (changes requested):
It is not as simple as just copying the text.

Now if we have a message with just an image (just an example, there're many edge cases), it will copy the filename of the image, not the image itself.

TheDeveloperRyan:
> It is not as simple as just copying the text.
> 
> Now if we have a message with just an image (just an example, they're many edge cases), it will copy the filename of the image, not the image itself.

The original intent of this change was to support copying text content only. Handling files and media introduces significantly more complexity — especially since clipboard support for different MIME types varies between browsers, which lead to inconsistent behavior or outright failures when copying media.

To keep the behavior predictable, we could disable the copy button entirely if the event type is not m.emote or m.text. That way we clearly scope the feature to text-based events only and avoid the edge cases around files and media for now.

TheDeveloperRyan:
Looking back at the original issue, I misunderstood the requirement — they were actually asking for an implementation that includes media content when copying a message, not just text.

To address this, I think the best approach would be to use ClipboardItem.supports() to check whether the media MIME type is supported by the user’s browser before attempting to copy it. If the MIME type isn’t supported, we can disable the copy button for that message.

I believe explicitly disabling the action is a better UX than allowing it to fail silently or throw an error, as it makes the limitation clear and avoids inconsistent behavior across browsers.

SpeeQz1:
@TheDeveloperRyan hello, has there been any additional work done on this PR. It would be great to have at minimum the ability to copy at least the text contents of a message.

Issue #2575, opened 2026-01-26
Display channel name in title


astranchet:
### Describe the problem

When I have several tabs opened, I don't know which on is which.

### Describe the solution you'd like

I would like the `title` tag to reflect the page I'm on : the name of a person, or of a canal

### Additional context

It's also an accessibility issue.

TheDeveloperRyan:
This would be a great addition so we can quickly identify our active rooms at a glance when multiple tabs are open. It’s more than just a convenience feature — dynamically updating the title is important for accessibility, as assistive tools rely on it to help users navigate efficiently.

I can put together a PR for this.

PR #2658, opened 2026-02-17
Add current room and space name to title

TheDeveloperRyan:
### Description
This PR adds a TitleUpdater component that changes the title of the browser tab dynamically based on the currently active room and space. With this feature, it is easier to distinguish between tabs when multiple rooms are open. This feature not only improves usability but also accessibility since assistive technologies depend on the correct title of the page to enable users to navigate through the page efficiently.

This addition was discussed in #2575 

Fixes #

#### Type of change

### Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas **( Code is self documenting )**

### DIFF
--- a/src/app/pages/client/ClientNonUIFeatures.tsx
+++ b/src/app/pages/client/ClientNonUIFeatures.tsx
@@ -8,7 +8,7 @@ import LogoUnreadSVG from '../../../../public/res/svg/cinny-unread.svg';
 import LogoHighlightSVG from '../../../../public/res/svg/cinny-highlight.svg';
 import NotificationSound from '../../../../public/sound/notification.ogg';
 import InviteSound from '../../../../public/sound/invite.ogg';
-import { notificationPermission, setFavicon } from '../../utils/dom';
+import { notificationPermission, setFavicon, setTitle } from '../../utils/dom';
 import { useSetting } from '../../state/hooks/settings';
 import { settingsAtom } from '../../state/settings';
 import { allInvitesAtom } from '../../state/room-list/inviteList';
@@ -26,6 +26,7 @@ import { getMxIdLocalPart, mxcUrlToHttp } from '../../utils/matrix';
 import { useSelectedRoom } from '../../hooks/router/useSelectedRoom';
 import { useInboxNotificationsSelected } from '../../hooks/router/useInbox';
 import { useMediaAuthentication } from '../../hooks/useMediaAuthentication';
+import { useSelectedSpace } from '../../hooks/router/useSelectedSpace';
 
 function SystemEmojiFeature() {
   const [twitterEmoji] = useSetting(settingsAtom, 'twitterEmoji');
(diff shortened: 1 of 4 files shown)


src/app/pages/client/ClientNonUIFeatures.tsx
```diff
@@ -76,6 +77,32 @@ function FaviconUpdater() {
...
+function TitleUpdater() {
+  const mx = useMatrixClient();
+  const selectedRoomId = useSelectedRoom();
+  const selectedSpaceId = useSelectedSpace();
+  
+  useEffect(() => {
+    let title = 'Cinny | ';
+    if (selectedSpaceId) {
+      const space = mx.getRoom(selectedSpaceId);
+      title += space?.name ?? 'Unknown Space';
```
mmmykhailo: I'm not a fan of the PR idea in general, but at least if you can't get the name - just don't include it in the title, "Unknown Space" is just useless for the user

mmmykhailo (changes requested):

src/app/pages/client/ClientNonUIFeatures.tsx
TheDeveloperRyan: Thanks for the feedback. I will update the implementation so that if the room or space name can’t be resolved, it’s omitted from the page title.

Regarding the PR itself, I understand that including this information in the page title might seem unnecessary or even undesirable from a UX perspective, as it may not look the best in your tab bar; however, it does provide real accessibility benefits laid out by [WCAG](https://www.w3.org/WAI/WCAG22/Understanding/page-titled.html)

If needed, we could make this configurable in settings, but I think due to the acceptability reason, it may be useful to keep it enabled by default.

mmmykhailo: Wdyt about making it the chat title first and then `Cinny`? Like `My room - My space - Cinny`
So that when a person listens to it they can quicker understand what chat they are in and skip the rest

src/app/pages/client/ClientNonUIFeatures.tsx
```diff
@@ -76,6 +77,32 @@ function FaviconUpdater() {
...
+    if (!roomName && !spaceName) {
+      title = 'Cinny';
+    }
+
+    setTitle(title);
+  }, [mx, selectedRoom, selectedSpace]);
+
+  return null;
+}
+
```
mmmykhailo:
```suggestion
function TitleUpdater() {
  const mx = useMatrixClient();
  const selectedSpace = useSelectedSpace();
  const selectedRoom = useSelectedRoom();

  useEffect(() => {
    const space = selectedSpace ? mx.getRoom(selectedSpace) : undefined;
    const room = selectedRoom ? mx.getRoom(selectedRoom) : undefined;

    const spaceName = space?.name || space?.roomId;
    const roomName = room?.name || room?.roomId;

    const parts: string[] = [];
    if (roomName) parts.push(roomName);
    if (spaceName) parts.push(spaceName);
    parts.push('Cinny');

    document.title = parts.join(' – ');
  }, [mx, selectedRoom, selectedSpace]);

  return null;
}
```

TheDeveloperRyan: Yeah, that actually is a better order for the title elements. I was basing my order on how Discord structures their title.

src/app/pages/client/ClientNonUIFeatures.tsx
TheDeveloperRyan: I actually do like this order better as it prioritizes the content over branding.

mmmykhailo (approved):

Issue #371, opened 2022-03-10
Collapse multiple events into a single line


N-R-K:
Currently all membership/nick events are listed out. This can be very annoying when using matrix-to-irc bridge where people leaving and joining is rather common. Hiding the events entirely is not desirable either since then you won't see the event at all.

Here's an example of how badly events can clutter the chat; IMO element's way of collapsing multiple events into a single line is a good way to handle things.

![image](https://user-images.githubusercontent.com/79544946/157618896-a452b1d7-6487-4dd4-9ebe-60ca0f568145.png)

N-R-K:
Any update on this? The problem still exists and is kinda annoying.

PR #2666, opened 2026-02-19
Feat: Collapse multiple events into a single line

Sollace:
### Description

This is my attempt at implementing the change for #371 (As it seemed like a reasonably easy thing for me to do). It's far from perfect but maybe it's enough to serve as a starting point for someone who wants to take this further and turn it into something more suitable for merging.

Essentially, with this patch any consecutive room member events (more than 3 in a row) that are also on the same day will be combined into a single message with a shortened summary of who joined, who left, and whether there were any non-join/leave events. Clicking on the message will reveal all of them so you can still access the entire timeline. Clicking again will also collapse them back down.

Edit: Updated screenshot of what they look like now:
image

I chose to lower the limit to 2 or more rather than 3.

There are a few issues that would need to be fixed before I can consider this truly "ready":
1. ~~The toggle triggers if you click anywhere in the message body, so if you try to open the options dropdown on one of the message it will toggle the collapse~~
2. ~~The chevron icon changes position when it's opened. It's centered vertically, and in general the whole styling of this UI element needs more work. (Maybe turn the chevron icon into a button?)~~
3. Pagination logic needs to be adjusted. Because we have multiple messages taking up one line, if you have a room that's just those messages it will end up continuing to load more of them until the screen is full. This might not be a major concern, I think it could result in more lag/server load from having to load more than the usual amount.
4. ~~Needs a user preference added to turn this behaviour on and off~~

#### Type of change

### Checklist:


Sollace:

Sollace:
Reworked the logic to avoid iterating through the events multiple times. The "Hide Membership Change" preference has also been renamed to "Membership Changes" with the options None, All, Summary (new default). When summary is selected it will do the collapsing, and None/All cover the two existing options of hiding and showing events without collapsing them.

image

Sollace:
A bit of a todo for myself. It looks like if messages aren't fully loaded for the day and it adds more, it only renders the new ones(?) and so doesn't roll them into the group that's already created.

image

I'm not sure how to properly fix that tbh

LinuxinaBit:
Is it possible to have this also apply to hidden events, profile changes, and room name/description changes?

Sollace:
> Is it possible to have this also apply to hidden events, profile changes, and room name/description changes?

Yeah absolutely. I don't know if I would include hidden events as you don't see them anyway, but the others I can totally do.

The question is do you want those rolled into the same groups, or have a group for room name/description changes on their own?

LinuxinaBit:
> I don't know if I would include hidden events as you don't see them anyway, but the others I can totally do.

It's an option in settings; General -> Messages -> Show Hidden Events
This screenshot has examples of profile changes, hidden events, and join/leave events:

image


> do you want those rolled into the same groups, or have a group for room name/description changes on their own?

It's hard to say, but I'd probably want them grouped together to avoid issues with alternating types of events.
The hard part there is figuring out what it should say when collapsed (assuming multiple things happened in the same time-span).
Maybe something like, "5 people joined, 2 people left, 1 person changed their username, and 4 hidden events occurred."

r-tae:
Element can do this, which gives a little more information than just counts

> Foo and 4 others left, Bar joined, changed their profile picture, and left, Baz joined

Sollace:
> > I don't know if I would include hidden events as you don't see them anyway, but the others I can totally do.
> 
> It's an option in settings; General -> Messages -> Show Hidden Events This screenshot has examples of profile changes, hidden events, and join/leave events:

Yeah I know, but it's not visible _by default_.

That said, I did change my mind and will probably have it combine them into "X hidden events"

> > do you want those rolled into the same groups, or have a group for room name/description changes on their own?
> 
> It's hard to say, but I'd probably want them grouped together to avoid issues with alternating types of events. The hard part there is figuring out what it should say when collapsed (assuming multiple things happened in the same time-span). Maybe something like, "5 people joined, 2 people left, 1 person changed their username, and 4 hidden events occurred."

That might work. I'll try some variants out when I next work on this.

What I'm thinking is I would only group hidden events with other hidden events.
Name/Profile changes can be easily added in with the leave/joins (as I believe they already do include invites/bans/kicks)

The other type of event to consider is changing the room name, icon, and topic which I would also combine into their own third group.

LinuxinaBit:
I like that concept too.

If possible, it would be nice to have some logic to combine them together out-of-order to take:
```
Alice joined
Bob changed room topic
Charlie left
```
and turn it into
```
Alice joined, Bob left
Bob changed room topic
```
while
```
Charlie joined
<Charlie> Hello everyone
Alice left
```
doesn't change because there's an actual message in-between.
This alternating event issue is why I personally prefer combining everything into one message.

For maximum compaction, I think it's a good idea to also combine events across dates, turning this:
image
into the single line:
`paulo89, weeeiiiiiiii, and 9 others joined, paulo89, camjerkforfeednl, and 1 others left`

Sollace:
> I like that concept too.
> 
> If possible, it would be nice to have some logic to combine them together out-of-order to take:
> 
> ```
> Alice joined
> Bob changed room topic
> Charlie left
> ```
> 
> and turn it into
> 
> ```
> Alice joined, Bob left
> Bob changed room topic
> ```
> 
> while
> 
> ```
> Charlie joined
> <Charlie> Hello everyone
> Alice left
> ```
> 
> doesn't change because there's an actual message in-between. This alternating event issue is why I personally prefer combining everything into one message.

So it already does this. They only combine when they're next to each other, so if there's an actual message in the middle the join and leave will be separate.

Edit: Oh okay, I think I see what you mean. - I don't know about whether we should be changing the order here as sometimes the order can be significant. In the case of the collapsed message it doesn't matter so much because the original order can be recovered by expanding the message to see all the original events.


> For maximum compaction, I think it's a good idea to also combine events across dates, turning this: 

The dates splitting is a bit of a compromise to deal with rooms that have only these types of events in them, as otherwise the timeline will never fill up and it will just continue to load the entire history.

Besides that, I do somewhat prefer to keep the date separation as a delineation between different days.

Sollace:
So I realised that I was already including profile changes in with the member events (and actually intended for them to be part of the summary) but they just weren't being because this whole time I'd been testing with profile changes hidden.

So now they are
image

I have also added the grouping requested for changes to the room name/topic/avatar and other room events that are hidden. It turned out easier to group them together with room state
image

There might be other "hidden" events that don't get grouped, but because of the way cinny considers any event it hasn't a special case for displaying as a "hidden" event I can't easily have something that groups all of them without _really_ digging deep into how it handles event types and I don't really want to wander too far.

PR #2664, opened 2026-02-19
Request session info from sw if missing

ajbura:
### Description

Fixes #2638

#### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings

PR #2670, opened 2026-02-20
Fix session fetching in service worker

mmmykhailo:
### Description

Resolves the issue found during review in #2664 

#### Known issues

- If you reload with clearing the cache it will still not load the image until reload, this is a limitation of how sw works and I don't think we can fix it with just sw changes

#### Type of change

### Checklist:

### DIFF
--- a/src/sw.ts
+++ b/src/sw.ts
@@ -104,20 +104,15 @@ self.addEventListener('message', (event: ExtendableMessageEvent) => {
   }
 });
 
-function isMediaRequest(url: string, baseUrl?: string): boolean {
-  const mediaPaths = ['/_matrix/client/v1/media/download', '/_matrix/client/v1/media/thumbnail'];
+const MEDIA_PATHS = ['/_matrix/client/v1/media/download', '/_matrix/client/v1/media/thumbnail'];
+
+function isMediaRequest(url: string): boolean {
   try {
     const { pathname } = new URL(url);
-    if (!mediaPaths.some((p) => pathname.startsWith(p))) return false;
+    return MEDIA_PATHS.some((p) => pathname.startsWith(p));
   } catch {
     return false;
   }
-  if (!baseUrl) return true;
-
-  const downloadUrl = new URL('/_matrix/client/v1/media/download', baseUrl);
-  const thumbnailUrl = new URL('/_matrix/client/v1/media/thumbnail', baseUrl);
-
-  return url.startsWith(downloadUrl.href) || url.startsWith(thumbnailUrl.href);
 }
 
 function fetchConfig(token: string): RequestInit {
@@ -131,7 +126,7 @@ function fetchConfig(token: string): RequestInit {
 
 function respondMediaRequest(event: FetchEvent, session: SessionInfo): void {
   const { url } = event.request;
-  if (!isMediaRequest(url, session.baseUrl)) return;
+  if (!isMediaRequest(url)) return;
 
   event.respondWith(fetch(url, fetchConfig(session.accessToken)));
 }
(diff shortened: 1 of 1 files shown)


mmmykhailo:

src/sw.ts
```diff
@@ -104,11 +104,15 @@ self.addEventListener('message', (event: ExtendableMessageEvent) => {
...
-  return url.startsWith(downloadUrl.href) || url.startsWith(thumbnailUrl.href);
+const MEDIA_PATHS = ['/_matrix/client/v1/media/download', '/_matrix/client/v1/media/thumbnail'];
+
+function isMediaRequest(url: string): boolean {
+  try {
+    const { pathname } = new URL(url);
+    return MEDIA_PATHS.some((p) => pathname.startsWith(p));
+  } catch {
+    return false;
+  }
```
ajbura, line 115: what does this change fixes?

mmmykhailo: There's no session here to get baseUrl, so I decided to remove it from the checker.
Image

ajbura: thats why it was called after having the session.

mmmykhailo: Yes, but this way we would delay anything, not just media

ajbura: this will only happen once and wont take long. we will receive message without session so its not like we are delaying indefinitely

src/sw.ts
```diff
@@ -139,10 +143,16 @@ self.addEventListener('fetch', (event: FetchEvent) => {
     return;
   }
 
-  const responsePending = requestSessionWithTimeout(clientId).then((s) => {
-    if (!s) return;
-    respondMediaRequest(event, s);
-  });
-
-  event.waitUntil(responsePending);
```
ajbura, line 147: we were already waiting for it here.

ajbura (commented):
would love to know what issue you found which this PR is fixing?

src/sw.ts
mmmykhailo: The core problem was that `respondWith` must be called synchronously.

The old code called `event.waitUntil()` with a promise that eventually called `respondMediaRequest`, which internally called `event.respondWith()` asynchronously and it was too late. This means the service worker's auth token injection was silently failing for media requests, because the browser had already dispatched the request without the access token.

`waitUntil` does keep the service worker alive until the promise resolves, but that's not the same as intercepting the request.

ajbura: sounds good. can you please just change just this part?

mmmykhailo: Sure, I'll revert the other changes 👍️

src/sw.ts
```diff
@@ -139,10 +143,16 @@ self.addEventListener('fetch', (event: FetchEvent) => {
...
+  const { url } = event.request;
+  if (!isMediaRequest(url)) return;
+
+  // respondWith must be called synchronously, so
+  // we pass a Promise and the browser
+  // suspends the request while we wait for the session
+  event.respondWith(
+    requestSessionWithTimeout(clientId).then((fetchedSession) => {
+      if (!fetchedSession) return fetch(url);
```
ajbura: we should do fetch(event.request) instead because we do not want to change headers and all?

ajbura (changes requested):

src/sw.ts
mmmykhailo: Actually, it seems to be an issue. The moment I remove this check a circular dependency happens: the SW is holding the page's JS/CSS waiting for the session, while the session can only be sent by the page's js, which can't run until those resources are fetched. The deadlock resolves only when the 3-second timeout fires.

Do you have other ideas on how we can avoid that? Or maybe I'm missing something?

ajbura: feel free to create a separate util to look for the media paths. but still keeps the one with baseurl for later use

src/sw.ts
mmmykhailo: What about this?
image

mmmykhailo: Since we think about initial headers we should merge them properly

ajbura (approved):
Overall looks good. just some minor tweaks are needed. I am merging it and will fix the remaining bits.

Thank you for the changes!

PR #2671, opened 2026-02-20
Add Support For Scheduled Messages via MSC4140

jasonlaguidice:
### Description

This PR adds the ability to schedule messages in rooms (both encrypted and otherwise) using the framework established in MSC4140 (cancellable delayed events).

I noticed that the API Cinny uses already has support for the unstable spec endpoints so this seemed like a pretty easy add.

image

This adds a chevron next to the send button which opens a simple date/time picker. When selected to schedule, the send icon changes to a clock and is given a color to indicate this will schedule the message instead of send per normal. The message is sent as a delayed event using MSC4140. There are some limitations to this - the API doesn't provide any method to decrypt these events so scheduled events in encrypted rooms can't be retrieved... only deleted. The other notable one is that the server has to support MSC4140 (and have a suitably large max allowable delay).

image

image

The chevron and options should only appear if the server reports that it supports MSC4140.

Scheduled messages in the selected room, if present, will display a chip with the count of scheduled messages. Clicking the chip will expand the selection, showing all scheduled messages in the room with the option to edit the message or cancel the scheduled send. As mentioned above - encrypted messages will NOT be previewed (and are just displayed as Encrypted Message) and there is no option to edit - only delete.

I have tested this against my homeserver and everything seems to be working as intended. But would like to open this up for others to help push buttons and find scenarios when this shouldn't work.

Future/pending changes to the Matrix API used should be simple fixes, if required at all.

Fixes #

#### Type of change

### Checklist:

### DIFF
--- a/src/app/features/room/RoomInput.tsx
+++ b/src/app/features/room/RoomInput.tsx
@@ -7,9 +7,10 @@ import React, {
   useRef,
   useState,
 } from 'react';
-import { useAtom, useAtomValue } from 'jotai';
+import { useAtom, useAtomValue, useSetAtom } from 'jotai';
 import { isKeyHotkey } from 'is-hotkey';
-import { EventType, IContent, MsgType, RelationType, Room } from 'matrix-js-sdk';
+import { EventType, IContent, MatrixError, MsgType, RelationType, Room } from 'matrix-js-sdk';
+import type { RoomMessageEventContent } from 'matrix-js-sdk/lib/@types/events';
 import { ReactEditor } from 'slate-react';
 import { Transforms, Editor } from 'slate';
 import {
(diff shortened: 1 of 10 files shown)


jasonlaguidice:

dollth-ing:

ajbura:
Thanks for the contribution! 🙌

For new features, we usually ask contributors to first open an issue or start a discussion so the maintainers can review the proposal and make sure it aligns with the project's roadmap.

Could you please create a feature request issue describing the motivation and proposed solution? Once the discussion is complete and the feature is approved, we’ll be happy to review the PR.

I’ll keep this PR on hold for now until the issue is created and discussed. Thanks again for your effort!

PR #2700, opened 2026-02-24
feature: display tag icons in members drawer

malwarepad:
### Description
Adds the tag icon (if one exists) on the left of the text in the members list. It is a small visual change but with enough power levels, it looks really really cool.

### Type of change

### Checklist:


malwarepad:

src/app/features/room/MembersDrawer.tsx
```diff
@@ -397,6 +406,7 @@ export function MembersDrawer({ room, members }: MembersDrawerProps) {
                         className={classNames(css.MembersGroupLabel, css.DrawerVirtualItem)}
                         size="L400"
                       >
+                        {tagIconSrc && <PowerIcon size="100" iconSrc={tagIconSrc} />}
```
kfiven, line 409: I think it looks much better if we display the tag after role name. Otherwise it feels like it's another user in there.

kfiven (changes requested):

src/app/features/room/MembersDrawer.tsx
malwarepad: > I think it looks much better if we display the tag after role name. Otherwise it feels like it's another user in there.

Even though that makes sense in a chat context (since its besides a user, and that user has a pfp), having the icon after the tag looks really bad on the sidebar. Other than role name width inconsistencies, it looks weird having the pfp & tag icons on opposite sides. Also there is a pretty big distinction between user pfps & tag icons, since the text size and icon sizes are vastly different.

malwarepad: @kfiven what do you think?

kfiven: Well.. tags inside room timeline also after username, also in settings they are after role name so I think a little consistency would be better.
@malwarepad Hi, how does it sounds?

malwarepad:
@kfiven Honestly, I tried putting it on the other side and it just looks bad. It makes sense in the chat context but not at all on the sidebar. No consistency is lost since its symmetrical if you think about it as well.
Screenshot_20260517_153013

PR #2706, opened 2026-02-28
feat: add the ability to reply with any filetype

nushea:
### Description

Whenever I talk with people and they ask me to send them a file or i want to reply with an image to something they say I currently have to reply with an empty message and then the image/file. 

This fix would allow anyone to reply with any file or a sticker to an event.

Also while you can send multiple files in one burst or a file and text, only the topmost item gets the reply event so it doesn't show as many reply icon visuals as there are events, with the reply working as before for text+file, with only the text having the reply tag as it is always the topmost item.

Additionally the older flow where you couldn't reply with images/files has confused all of the people that i have on-boarded or that were new to the platform

Example of how this feature could be used
image

Fixes #2659 

#### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #2708, opened 2026-02-28
fix: reply properly when attachments are present

malwarepad:
### Description
When a reply is in queue and the user sends an attachment, mark the new message as a reply. 

I also noticed that captions aren't used (like for example in Element X), but I presume that is the case in order for bulk uploading to be cleaner. Hence I didn't mess with that at all.

#### Type of change

### Checklist:


nushea:
I think this fixes a subset of the issues that are being resolved by #2706

malwarepad:
> I think this fixes a subset of the issues that are being resolved by #2706

Yup you're right, I didn't check existing PRs since I had noticed that bug for quite a while. You even fixed the sticker thing, great job

PR #2713, opened 2026-03-01
Feat: Uniform avatar appearance in space/room navigation

wolterkam:
Introduces a new appearance setting turning room avatar icons more consistent, reducing the visual noise caused by mismatched shapes and styles.

### Problem
In the left sidebar navigation between spaces and rooms, some avatar tiles filled their button area while others appeared as small icons or irregular transparent shapes. There is no canon style to how these avatars appear, so visual variety is expected, but the stark differences in how avatars occupied their space made the sidebar feel inconsistent with the rest of the design language. 

### Solution
Avatar icons can now fill their background with the dominant color of the icon itself, harmonizing the appearance of otherwise mismatched tiles. This can be toggled via Consistent Icon Style in Appearance settings. I've shipped it disabled by default, but happy to change that if there's a preference either way.

<table>
  <tr>
    <td>icons2</td>
    <td>icons1</td>
  </tr>
</table>

### Type of change

### Checklist:

### DIFF
--- a/src/app/components/room-avatar/AvatarImage.tsx
+++ b/src/app/components/room-avatar/AvatarImage.tsx
@@ -1,0 +1,40 @@
+import { AvatarImage as FoldsAvatarImage } from 'folds';
+import React, { ReactEventHandler, useState } from 'react';
+import { useSetting } from '../../state/hooks/settings';
+import { settingsAtom } from '../../state/settings';
+import bgColorImg from '../../../util/bgColorImg';
+import * as css from './RoomAvatar.css';
+
+type AvatarImageProps = {
+  src: string;
+  alt?: string;
+  uniformIcons?: boolean;
+  onError: () => void;
+};
+
+export function AvatarImage({ src, alt, uniformIcons, onError }: AvatarImageProps) {
+  const [uniformIconsSetting] = useSetting(settingsAtom, 'uniformIcons');
+  const [image, setImage] = useState<HTMLImageElement | undefined>(undefined);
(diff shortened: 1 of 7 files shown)


7w1:
You need to add `crossOrigin="anonymous"` into the FoldsAvatarImage in AvatarImage.tsx or this will immediately crash all Firefox based browsers for whatever reason.

wolterkam:
> You need to add `crossOrigin="anonymous"` into the FoldsAvatarImage in AvatarImage.tsx or this will immediately crash all Firefox based browsers for whatever reason.

Thank you, I've implemented a fix. In short: Firefox blocks pixel access from cross-origin sources by default, but that restriction doesn't make sense for our use case, so the recommended workaround is safe here.

kfiven (commented):
I feel this is too much of control on user input. I can see down the road people complaining where does the background of their images came from.

wolterkam:
> I feel this is too much of control on user input. I can see down the road people complaining where does the background of their images came from.

Hmm... Operating systems from across the board have been doing this for years without much backlash. 

It's a setting, off by default, builds off the existing design, and addresses a blind spot.

The same visual component is a delineated rounded rectangle in every other case:
- Action button
- Rectangular avatar image
- Auto-created "initials" buttons

My goal was to conform the outlier to a design decision that is already established and uncontroversial.

kfiven:
I am all in for streamlined UI but I think we should only do that on UI component and User Input such as background should not be done. Also I think one of our core philosophy is to only have necessary settings which keeps the app including settings simple.

PR #2715, opened 2026-03-02
feat: add the ability to reply with any filetype

### LINKED ISSUE #2659: Selecting "reply" on a message and then sending an image uploads the image without replying
### Describe the bug

Currently you can't reply to a message with just an image upload. Hitting enter on the chatbox will cause cinny to send the image without attaching it to the message your replying to causing you to look like a crazy person.

Then to send the actual reply you have to send an empty message with something like "." that then puts the reply link below the image, possible causing confusion.

### Reproduction

1. Find a image online, any image. Right-click > copy image (to put it on your clipboard)
2. Within cinny, click the "reply" button next to an image so the message box has a reply queued up
3. Press Ctrl+V to paste the image from the clipboard
4. Wait for the send button to appear on the image
5. Click on the message box and press Enter.

### Expected behavior

The message with the image attached should be made in reply to the message you had selected earlier.

### Actual Behaviour

The image is sent without the reply. The reply remains selected on the message box waiting for a message to be input. If you enter a message after that and press enter, it will then send your message as a reply but underneath the already posted image.

### Platform and versions

```shell
1. Manjaro
2. Floorp
3. v4.10.3 (app.cinny.in)
```

### Additional context

N/A

nushea:
### Description
Whenever I talk with people and they ask me to send them a file or i want to reply with an image to something they say I currently have to reply with an empty message and then the image/file.

This fix would allow anyone to reply with any file or a sticker to an event.

Also while you can send multiple files in one burst or a file and text, only the topmost item gets the reply event so it doesn't show as many reply icon visuals as there are events, with the reply working as before for text+file, with only the text having the reply tag as it is always the topmost item.

Additionally the older flow where you couldn't reply with images/files has confused all of the people that i have on-boarded or that were new to the platform

Example of how this feature could be used
image

Fixes #2659
(accidentally closed https://github.com/cinnyapp/cinny/pull/2706 by renaming the branch to give my local branches proper names ~-~ without realizing it automatically closes them on github/cinny)

#### Type of change

### Checklist:

### DIFF
--- a/src/app/features/room/RoomInput.tsx
+++ b/src/app/features/room/RoomInput.tsx
@@ -298,6 +298,7 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
     };
 
     const handleSendUpload = async (uploads: UploadSuccess[]) => {
+      const plaintext = toPlainText(editor.children, isMarkdown).trim();
       const contentsPromises = uploads.map(async (upload) => {
         const fileItem = selectedFiles.find((f) => f.file === upload.file);
         if (!fileItem) throw new Error('Broken upload');
@@ -317,7 +318,6 @@ export const RoomInput = forwardRef<HTMLDivElement, RoomInputProps>(
       const contents = fulfilledPromiseSettledResult(await Promise.allSettled(contentsPromises));
 
       if (contents.length > 0) {
-        const plaintext = toPlainText(editor.children, isMarkdown).trim();
         const replyContent = plaintext.length === 0 ? getReplyContent(replyDraft) : undefined;
         if (replyContent) contents[0]['m.relates_to'] = replyContent;
         setReplyDraft(undefined);
--- a/src/app/features/room/RoomInput.tsx
+++ b/src/app/features/room/RoomInput.tsx
@@ -9,7 +9,7 @@ import React, {
 } from 'react';
 import { useAtom, useAtomValue } from 'jotai';
 import { isKeyHotkey } from 'is-hotkey';
-import { EventType, IContent, MsgType, RelationType, Room } from 'matrix-js-sdk';
+import { EventType, IContent, IEventRelation, MsgType, RelationType, Room } from 'matrix-js-sdk';
 import { ReactEditor } from 'slate-react';
 import { Transforms, Editor } from 'slate';
 import {
@@ -68,6 +68,7 @@ import { useFilePicker } from '../../hooks/useFilePicker';
 import { useFilePasteHandler } from '../../hooks/useFilePasteHandler';
 import { useFileDropZone } from '../../hooks/useFileDrop';
 import {
+  IReplyDraft,
   TUploadItem,
   TUploadMetadata,
   roomIdToMsgDraftAtomFamily,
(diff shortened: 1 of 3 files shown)


malwarepad (approved):

Sollace:
Found a bug!
If you add an attachment and type a message, both the message and the attachment generate a reply.

Ideally if there's message text the reply should only be applied to that.

image

nushea:
> Found a bug! If you add an attachment and type a message, both the message and the attachment generate a reply.
> 
> Ideally if there's message text the reply should only be applied to that.
> image

the intended behavior is for it to only apply to the topmost item (ie the message), but the pr is rather old now, maybe something broke in a later version if you have merged it to it? or honestly the implementation is a bit hacky because the entire message sending flow is quite hacky too, so if the message sends fast enough (kinda impossible tbh in my exp) it could be doing it so fast it clears the editor before the file sending has a chance to review it, i could update it to work better if this is reproduceable

Sollace:
> maybe something broke in a later version if you have merged it to it

Dunno. That might be the case but I also don't remember exactly when I merged it to my branch. I don't think I've pulled since this was last reviewed though.

nushea:
> > maybe something broke in a later version if you have merged it to it
> 
> Dunno. That might be the case but I also don't remember exactly when I merged it to my branch. I don't think I've pulled since this was last viewed though.

oh sorry i just realized that this was still using an older implementation of it, the issue is that cinny sends files and text in parallel so there can be a race condition where the editor gets cleared out before you get the chance to send the files, i have made a one line change to fix this by literally beating out the clear by any means possible, honestly the whole message sending flow could use changing but for this one option it might not be worth it

also there is another way to make it consistent but that would require awaiting the completion of creating the file contents which could make sending them a bit slower

Could you please update it on your fork (or move the plaintext function in the /src/app/features/room/RoomInput.tsx → handleSendUpload function to have const plaintext = toPlainText(editor.children, isMarkdown).trim(); as the first item ) and then tell me if the issue persists?

Sollace:
> > > maybe something broke in a later version if you have merged it to it
> > 
> > 
> > Dunno. That might be the case but I also don't remember exactly when I merged it to my branch. I don't think I've pulled since this was last viewed though.
> 
> oh sorry i just realized that this was still using an older implementation of it, the issue is that cinny sends files and text in parallel so there can be a race condition where the editor gets cleared out before you get the chance to send the files, i have made a one line change to fix this by literally beating out the clear by any means possible, honestly the whole message sending flow could use changing but for this one option it might not be worth it
> 
> also there is another way to make it consistent but that would require awaiting the completion of creating the file contents which could make sending them a bit slower
> 
> Could you please update it on your fork (or move the plaintext function in the /src/app/features/room/RoomInput.tsx → handleSendUpload function to have const plaintext = toPlainText(editor.children, isMarkdown).trim(); as the first item ) and then tell me if the issue persists?

Alright, pulled your branch and it looks to be good now!

Tested also after merging to my main branch and that has definitely solved the issue.

> honestly the whole message sending flow could use changing but for this one option it might not be worth it

Yeah by the sounds of it, that would probably be a good idea. Maybe for a future PR though.

Thanks for the speedy fix!

PR #2738, opened 2026-03-07
Move the message context outside of message content

smokku:
This change moves the message context (i.e. reply indicator) outside of actual message content.

### Modern layout

image

image

image

### Bubble layout

image

image

image

### Compact layout

image

image

image


ajbura:
Closing as not planned

PR #2740, opened 2026-03-07
Change play icon and add clickable backdrop on embedded videos.

Johnnycyan:
### Description
This is a small change that just updates the icon displayed over embedded videos to be more minimal, with a slight hover animation, and allows the user to click the entire video embed area to initiate playback instead of just the button itself.

If you would prefer to keep the old button style but still add the backdrop clickable area, I could modify the change to do that. I just felt as though the button didn't really look right.

Comments weren't necessary as this is a very small change.

My reason for making this change is because when I was using Cinny and a friend sent me a video, the first thing I tried to do was just click the video area, and nothing happened. This initially led me to believe I had to download the video to watch it as I didn't notice the "Watch" button initially.

https://github.com/user-attachments/assets/12155754-05d1-4da6-a612-33fadc3f0603

#### Type of change

### Checklist:

### DIFF
--- a/src/app/components/message/content/VideoContent.tsx
+++ b/src/app/components/message/content/VideoContent.tsx
@@ -130,17 +130,15 @@ export const VideoContent = as<'div', VideoContentProps>(
           </Box>
         )}
         {!autoPlay && !blurred && srcState.status === AsyncStatus.Idle && (
-          <Box className={css.AbsoluteContainer} alignItems="Center" justifyContent="Center">
-            <Button
-              variant="Secondary"
-              fill="Solid"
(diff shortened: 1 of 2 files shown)


Johnnycyan:

ajbura (changes requested):
i would like to keep the watch button as change makes it keyboard unfriendly and also does not follow design system. 

a better alternative could be to just add the onClick and onKeyDown to the outer `AbsoluteContainer`. 

also i think the problem is little more complex as the video can be spoiler also and same UX also need to be maintained with the Image message

PR #1576, opened 2023-12-08
Add translation support using i18next

aceArt-GmbH:
### Description

Minimal version of #681.
Allows follow up PRs which add languages, translate components, [language switcher](https://locize.com/blog/react-i18next/#language-switcher) or translator platform.

Fixes #164

Following https://locize.com/blog/react-i18next/

#### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #2747, opened 2026-03-09
feat: add i18n for most pages

weareoutman:
### Description

This PR adds i18n for almost all pages, while #1576 introduced i18next but never progressed on again.

#### Type of change

### Checklist:


weareoutman:

weareoutman:
It is not fully done yet, and it evolves lots of file changes, let me know if this contribution makes sense for you, or how can I make it more reviewable.

ajbura:
Hey @weareoutman, it largely looks good. can we have an option in settings to change the language to test so we can switch it without totally depending on system wide language change. And also can track the progress of what has/not been translated in issue

weareoutman:
Added the language option in settings > general:

image

Except for this, it should only contains translation changes, and now translation should cover most pages.

weareoutman:
Rebased and ready for review 🍻

Stealthii:
> Added the language option in settings > general:
> image
> 
> Except for this, it should only contains translation changes, and now translation should cover most pages.

top option should be System setting, and the default.

There's also some odd decisions about fields to translate, such as `dialog.userIdPlaceholder` (which was "@username:server", but now "@用户名:服务器" in your zh changes.

Did you perform these translations yourself, or use assistive tooling?

Stealthii:
@weareoutman Personal opinion here - I would separate the implementation of i18n support from a PR with your translations to another language. These are two separate things to review and fall under different scrutiny, besides until it's decided what should be translated, performing translations may waste work.

weareoutman:
> top option should be System setting, and the default.

From my experience, most websites and apps don't offer the 'System setting' option for language.

Edit: `i18next-browser-languagedetector` will make it auto select the actual system setting language at initial.

> There's also some odd decisions about fields to translate, such as dialog.userIdPlaceholder (which was "@username:server", but now "@用户名:服务器" in your zh changes.

It's a placeholder, so I think translation here is reasonable.

> Did you perform these translations yourself, or use assistive tooling?

Yes, changes are massive, so I have to use some tools, but all with careful manual review and corrections.

>  I would separate the implementation of i18n support from a PR with your translations to another language.

I agree with that, but the initial i18n support was already done in #1576 , this PR only introduces translations and a language selection in settings panel.

@Stealthii Thanks for your advice.

weareoutman:
@ajbura Congrats on your newly release which supports Voice/Video calls, that's a great work.

Do you have time to review this? I'll rebase it once you'd like to review, since the chance of conflicts is high for each new commits on the main branch.

ajbura:
@weareoutman Thanks you for the support!

While it looks good overall. I can not review the zh.json file as I do not know the language. As said by other contributor I also think it would be a good idea to have zh change as a new PR so we can ask other community member who are speaker of language for review without blocking the translation support changes.

kfiven:
We can setup the translation over Weblate, so I think we should just do the sting(-ification) here.

weareoutman:
@ajbura That makes sense. Rebased and removed zh.json, also added a 'supportedLanguages' in config.json to enable language selection in settings panel.

For my situation, manually adding zh.json and editing config.json after build would do the trick too, even if `zh.json` is not included by default (same to other languages).

@kfiven Not farmiliar with Weblate, but I guess it can help to generate `locales/*.json` based on `locales/en.json*.

config.json
```diff
@@ -27,5 +27,7 @@
   "hashRouter": {
     "enabled": false,
     "basename": "/"
-  }
+  },
+
+  "supportedLanguages": [{ "code": "en", "name": "English" }]
```
kfiven: Hi, thank you for changes. I do not think it is a good idea to change language based on config.

Language is a personal choice should be configurable by user or in future we can read system language and set that automatically.

kfiven (changes requested):

config.json
weareoutman: No, it's not for changing language, but for enabling the language selection in settings panel.

E.g. I can edit config.json with:

```json
  "supportedLanguages": [{ "code": "en", "name": "English" },{ "code": "zh", "name": "简体中文" }]
```

So that there will be a language selection with those two options in settings panel.

Maybe we can change the var name so that it is not confusing users, such as `languageOptions`?

weareoutman: There have to be a config so that users (self hosting) can extend language options without adding every locale translations in official cinny (they can add `dist/public/locales/*.json` as they want)

weareoutman: See changes of `src/app/features/settings/general/General.tsx`:

```js
  if ((supportedLanguages?.length ?? 1) <= 1) {
    // Hide the language setting if there is only one or no supported languages
    return null;
  }
```

kfiven: > So that there will be a language selection with those two options in settings panel.

I am still not sure why this is needed. The language selection in settings should show options based on what locale files are present. This way we do not have to add config for each language we want to add.

weareoutman: @kfiven This will allow self-hosting users add their own new languages whatever they want by:

- Download Cinny official tarball
- Put their locales in `dist/public/locales` such as `zh.json`, `jp.json`
- Add language options in `dist/config.json`

Because Cinny is pure static, it can not see what locale files are present at runtime.

weareoutman: I can move the language selection part out of it with another separated PR if you'd like to.

kfiven: That would be good. Otherwise we would be blocking this.

kfiven:
> @kfiven Not farmiliar with Weblate, but I guess it can help to generate locales/*.json based on `locales/en.json*.

It's a web based platform where translators can go and translate in the language they want. Then bot sync them back to the repo.

config.json
weareoutman: Removed the language selection and reverted config.json

PR #2795, opened 2026-03-17
fix: set body background color

### LINKED ISSUE #2527: White area near the keyboard on iOS
### Describe the bug

There’s a white area above the keyboard when I use Cinny as a web app on iOS.

https://github.com/user-attachments/assets/11ca16db-f584-41d4-a9a0-42f92e13fb59

The white corners at the bottom are not actually visible (they’re there because iPhones have round edges), but the white area above the keyboard is visible. 
It doesn’t appear in Telegram, or when writing posts on Mastodon or Discourse, so I believe this is not a general issue with iOS web apps.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
iOS 26.1
Safari 26.1
Cinny v4.10.1
```

### Additional context

_No response_

weareoutman:
### Description

Set body background to `color.Background.Container` to remove the white edges on iOS in dark theme.

Before:
Cinny-before

After:
Cinny-after

Fixes #2527

#### Type of change

### Checklist:

### DIFF
--- a/src/app/pages/App.css.ts
+++ b/src/app/pages/App.css.ts
@@ -1,0 +1,6 @@
+import { globalStyle } from '@vanilla-extract/css';
+import { color } from 'folds';
+
+globalStyle('body', {
+  backgroundColor: color.Background.Container,
+});
--- a/src/app/pages/App.tsx
+++ b/src/app/pages/App.tsx
@@ -12,6 +12,7 @@ import { FeatureCheck } from './FeatureCheck';
 import { createRouter } from './Router';
 import { ScreenSizeProvider, useScreenSize } from '../hooks/useScreenSize';
 import { useCompositionEndTracking } from '../hooks/useComposingCheck';
+import './App.css';
 
 const queryClient = new QueryClient();
 


u9000-Nine (approved):

src/app/components/emoji-board/components/styles.css.ts
```diff
@@ -1,6 +1,10 @@
-import { style } from '@vanilla-extract/css';
+import { style, globalStyle } from '@vanilla-extract/css';
 import { toRem, color, config, DefaultReset, FocusOutline } from 'folds';
 
+globalStyle('body', {
+  backgroundColor: color.Background.Container,
+});
+
```
ajbura: not sure why this is added to this file? isn't index.css a good fit for the change

ajbura (changes requested):

src/app/components/emoji-board/components/styles.css.ts
weareoutman: Oops, that's a mistake.

Added a new `App.css.ts` (index.css would not work, because it uses @vanilla-extract/css and color from folds)

PR #2798, opened 2026-03-18
fix: do not attempt to join call on doubleclick if missing permissions

YoJames2019:
### Description

Previously, if you double clicked a vc channel without the proper permissions it would attempt to join the call but obviously would not be able to, so it would show you a generic element call error screen and you would be stuck in that call unable to leave or join any other call until you refreshed the page. 

Now it does not attempt to start/join the call if you do not have permission.

#### Type of change

### DIFF
--- a/src/app/features/room-nav/RoomNavItem.tsx
+++ b/src/app/features/room-nav/RoomNavItem.tsx
@@ -23,12 +23,12 @@ import { useAtom, useAtomValue } from 'jotai';
 import { NavItem, NavItemContent, NavItemOptions, NavLink } from '../../components/nav';
 import { UnreadBadge, UnreadBadgeCenter } from '../../components/unread-badge';
 import { RoomAvatar, RoomIcon } from '../../components/room-avatar';
-import { getDirectRoomAvatarUrl, getRoomAvatarUrl } from '../../utils/room';
+import { getDirectRoomAvatarUrl, getRoomAvatarUrl, getStateEvent } from '../../utils/room';
 import { nameInitials } from '../../utils/common';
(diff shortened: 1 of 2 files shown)


YoJames2019:

src/app/features/room-nav/RoomNavItem.tsx
```diff
@@ -286,9 +287,14 @@ export function RoomNavItem({
   const callPref = useAtomValue(useCallPreferencesAtom());
   const autoDiscoveryInfo = useAutoDiscoveryInfo();
 
+  const powerLevels = usePowerLevels(room);
+  const creators = useRoomCreators(room);
+  const permissions = useRoomPermissions(creators, powerLevels);
+  const hasCallPermission = permissions.event(StateEvent.GroupCallMemberPrefix, mx.getSafeUserId());
```
ajbura: i really think that reading permission in reactive way is not necessary because our view does not need this info. we can save time of computation if we read this permission statically in click handler.

let me know how you think about it?

ajbura (changes requested):

src/app/features/room-nav/RoomNavItem.tsx
YoJames2019: i agree and I've made some of the changes, but I cannot find a way to read powerlevels for a room statically currently? Is there something I'm missing?

ajbura: you can use export `getPowersLevelFromMatrixEvent` from usePowerLevels and can use `readPowerLevel: ReadPowerLevelAPI`

ajbura (approved):
Looks great!

PR #2830, opened 2026-03-22
feat: enable Capacitor support for cinny

ntodten:
Added configuration for running cinny web app on iOS via capacitor. Ran on iOS 26.2 iPhone 17 simulator and real iPhone 15 iOS 26.3.1. Fixed a couple issues:
- Images, profile pictures, and room photos not loading
    - WKWebView (aka Capacitor runtime) does not allow sending Authorization headers in  tags resulting in 401s when fetching media
    - Implement a function fetching the image with the auth token and providing a local URL to the final resouce caller
- Theme backgrounds not filling safe/curved screen regions
    - Allow the background to be set by the theme instead of hard coding it

Note: The app can sometimes take up to 30s to establish a connection when opening or after being in the background for a while.


ntodten:

kfiven:
Hi, Thank you for this! Unfortunately there's little value in wrappers instead of PWAs which provides the same functionality on iOS.

I am closing as we do not plan to maintain and ship capacitor based app yet.

Issue #1569, opened 2023-12-02
Replies to spoilers reveal spoilers


razetime:
### Describe the bug

When replying to a message with a spoiler it is shown as raw text.

### Reproduction

1. type a spoilered message `||spoiler||`
2. reply to it
3. You will see spoiler contents

### Expected behavior

Spoiler must not be visible in the reply

### Platform and versions

```shell
1. OS: Devuan Daedalus
2. Browser: Firefox 115.4.0esr (64-bit)
3. Cinny version: v3.2.0 (app.cinny.in)
4. Matrix Homeserver: coven.palaiologos.rocks
```

### Additional context

![image](https://github.com/cinnyapp/cinny/assets/11015319/3d77b0ae-8015-4885-a707-0dd87759f6a4)

aceArt-GmbH:
replies do not render markdown at all

wyattwas:
I am new to the project and would like to contribute to it. I think i found the right spot but could use some help with it. If someone would be up for working on this, I'd be glad to help.

u9000-Nine:
> replies do not render markdown at all

This is true, but they really should for spoilers. Not doing so defeats the point of using spoilers at all.

u9000-Nine:
> I am new to the project and would like to contribute to it. I think i found the right spot but could use some help with it. If someone would be up for working on this, I'd be glad to help.

Sure! Join us in [#cinny:matrix.org](https://matrix.org/#cinny:matrix.org), and then it's the "Development" room.

Thanks for contributing 🙂

PR #2840, opened 2026-03-24
fix: Render formatted body in reply and editor

wyattwas:
This PR fixes issue #1569 by using the formatted_body of the message being replied to. It passes that to the HTMLReactParser. This approach not only renders spoilers, but all markdown/HTML that is contained in the formatted_body. It removes the `<br/>` tag, in order to show the entire message in one line.
I am definitely open for suggestion on how to improve this.

Changes:
- use the formatted_body of the message being replied to, if that message is formatted
- also use the formatted_body in the editor

Commits:
- render formatted body in reply and editor


wyattwas:

Issue #2619, opened 2026-02-13
Add ability to change the font


dembim:
### Describe the problem

The font bundled with Cinny does not have support for many of the characters I need. This makes it impossible or ugly to type in other languages, math expressions, etc.

### Describe the solution you'd like

I would like to be able to specify the chat/UI typeface in the appearance settings, to select from any system fonts I have installed on my computer.

u9000-Nine:
See also [Add support of Custom themes · Issue #357 · cinnyapp/cinny](https://github.com/cinnyapp/cinny/issues/357).

This could reasonably be argued to be a separate feature from custom themes though due to localization issues.

dembim:
This issue is definitely distinct from #357 (and higher priority, in my opinion). The requirement for user-configurable fonts stems from accessibility and inclusion concerns, whereas #357 appears to my motivated by aesthetics— user-configurable font support allows setting accessible/legible fonts for low-vision users, dyslexia-friendly fonts, fonts for non-latin scripts (e.g. for communities of speakers of non-latin-script languages), and so forth.

Alen21137:
Adding this feature would make Cinny far more accessible and user-friendly.

PR #2841, opened 2026-03-24
feat: add Vazirmatn fallback font for Arabic/Persian

ammahmoudi:
Bundles Vazirmatn (WOFF2) in public/font/vazirmatn/ to avoid runtime external font requests.
Adds @font-face with Arabic unicode ranges and includes OFL.txt.
Updates the global font stack to use Vazirmatn as a fallback after Inter, so Arabic/Persian text renders more cleanly without changing the UI font for everyone.


public/font/vazirmatn/OFL.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2015 The Vazirmatn Project Authors (https://github.com/rastikerdar/vazirmatn)
```
kfiven, line 1: Hi, what is the issue you are having?

ammahmoudi: Hi! The current default font doesn't render Persian and Arabic characters optimally, which affects legibility. Adding Vazirmatn as a fallback significantly improves readability and typographic balance for native readers, providing a much better user experience for these languages. Vazirmatn is common in apps with nice support for rtl users.

u9000-Nine: See also [Add ability to change the font · Issue #2619 · cinnyapp/cinny](https://github.com/cinnyapp/cinny/issues/2619) ([Farside](https://farside.link/https://github.com/cinnyapp/cinny/issues/2619))

dembim:
In general, I don't think adding a new fallback font for every script that renders incorrectly (which is most of them at the moment, as far as I can tell) is a scalable solution. A more broadly suitable approach is to allow in-client configuration of fonts, including system-provided fonts, so that rendering concerns for non-latin scripts can be handled on an individual basis. See the above-linked issue on adding such support.

ammahmoudi:
> In general, I don't think adding a new fallback font for every script that renders incorrectly (which is most of them at the moment, as far as I can tell) is a scalable solution. A more broadly suitable approach is to allow in-client configuration of fonts, including system-provided fonts, so that rendering concerns for non-latin scripts can be handled on an individual basis. See the above-linked issue on adding such support.

I agree that user-configurable fonts (#2619) is valuable, but it doesn’t fully solve this problem in practice for Arabic/Persian:

- Many users (especially on mobile) can’t realistically install additional system fonts, so “pick a system font” often won’t include something like Vazirmatn.
- A lot of real usage is mixed English + Persian/Arabic. A good fallback stack is the common approach in chat apps (Telegram-style), because it preserves the Latin UI typography while improving Arabic-script readability.
- This PR is intentionally narrow: it adds a single Arabic-script fallback via `unicode-range`, so it doesn’t change Latin UI rendering and doesn’t attempt to add “a font per script”.

If maintainers still prefer “config-only”, I understand. But I think a targeted Arabic-script fallback is a practical baseline improvement for a large user group, while font configuration can remain an additional power-user feature.

Alen21137:
This feature would significantly enhance the experience for Arabic/Persian users and other communities using RTL languages.

Behzad9 (approved):
I believe this feature would significantly improve readability and typographic balance for RTL users.

Alen21137 (approved):
I have also requested this feature in several other places, and it could be very beneficial for all Persian/Arabic users.

public/font/vazirmatn/OFL.txt
kfiven: Thoughts on https://fonts.google.com/noto ?

I myself is multilingual and looking at noto, they support wide variety of languages, which would be better approach rather than adding font for each language.

Behzad9: > Thoughts on https://fonts.google.com/noto ?
> 
> I myself is multilingual and looking at noto, they support wide variety of languages, which would be better approach rather than adding font for each language.

While it is true that noto fonts provide broad multilingual coverage, Vazirmatn is specifically optimized for Persian and other RTL languages. It is widely used in RTL contexts and is currently the default Persian font in X, reflecting its strong adoption.

PR #2851, opened 2026-03-25
fix: room still appearing after leaving it

### LINKED ISSUE #1623: Removed room/space doesn't disappear until 1 or 2 page refreshes
### Describe the bug

When I remove a room or a space, it doesn't optimistically disappear. It does so only after first or second page refresh.

### Reproduction

1. Go to spaces/rooms list
2. Right click on one and choose "Leave"

### Expected behavior

The space/room should disappear optimistically.

### Platform and versions

```shell
1. OS: MacOS Sonoma 14.1.1
2. Browser: Chrome, version 120.0.6099.199 (Official Build) (arm64)
3. Cinny version: 3.2.0
4. Matrix homeserver: matrix.org
```

### Additional context

_No response_

CrazyNicc:
Fixes: #1623 

After leaving a Room, the room would still show up in the room List, only disappearing after a "Cache Clear and Reload". This seems to be because the matrix-js-sdk still caches the old room membership.

This is sort of a hacky workaround, but it fixes the problem.

I have seen the Notice that you do not accept any new PRs until the SDK has been replaced, so this is in hopes of a quick fix making it through, or atleast of a documented fix for the problem that people can patch their local Cinny deployment with.

Disclaimer: I have used AI assistance to help me find the relevant file to edit, help me troubleshoot and supply the fix.

### DIFF
--- a/src/app/components/leave-room-prompt/LeaveRoomPrompt.tsx
+++ b/src/app/components/leave-room-prompt/LeaveRoomPrompt.tsx
@@ -28,12 +28,11 @@ type LeaveRoomPromptProps = {
 };
 export function LeaveRoomPrompt({ roomId, onDone, onCancel }: LeaveRoomPromptProps) {
   const mx = useMatrixClient();
-  const setRoomsAtom = useSetAtom(allRoomsAtom);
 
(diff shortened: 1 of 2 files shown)


CrazyNicc:

Issue #357, opened 2022-03-06
Add support of Custom themes


t3dium:
An element style dark mode would be nice since I think the current dark modes colour scheme looks a bit outdated.

Can we have a theme with: text background #15191E, sidebar #23272C, and the people+home sidebar to be #2A282B

Also, an option to change font size would be nice, by default the font looks quite massive

ghost:
It would be better to have the custom theme feature instead of asking for a theme to be added to Cinny.

GTcreyon:
How exactly would this be implemented? My understanding of Tauri is limited, but I'm not aware of any way to import custom CSS from the filesystem without CORS issues.

Kommynct:
You could cheat and use the element themes backend, and just assign the css selectors to the same names as element.

This would also make the themes interoperable, and bring all of the element themes to cinny.

archeite:
> How exactly would this be implemented? My understanding of Tauri is limited, but I'm not aware of any way to import custom CSS from the filesystem without CORS issues.

Maybe a text box to paste CSS into? Or maybe how Element (Labs) does it with theme URLs? (https://github.com/vector-im/element-web/blob/master/docs/theming.md)

LilyLavenderPony:
Hey, is there still interest in custom theming for Cinny?

I have been looking into this and would be interested in working on a lightweight approach, but I want to check first if this aligns with the maintainers’ direction.

Instead of a full theme system, I would

- introduce a few "stable hooks" (class names) in key layout components so styling doesn’t depend on hashed Vanilla Extract classes
- make certain foreground layers partially transparent, so a user-defined background can show through
- expose simple user options to set a linear gradient background and angle.

The goal is to enable some level of runtime customisation without refacturing the current styling system or requiring theme import/export.

From what I have seen, working purely with external CSS is quite fragile due to hashed class and variable names, so this seems like a feasable way to open up basic theming.

To the maintainers: Does this approach sound acceptable?

u9000-Nine:
From my conversations with @kfiven, that seems like the direction he was planning.

I'd want the transparency to be definable though, since that can affect accessibility. Also a blur setting would be nice if we're going to implement transparency

LilyLavenderPony:
Thanks, I agree. I’m working on this now and will open a pull request once it's in a good state.

Alen21137:
Persian (Farsi) and other RTL languages are not properly supported in Cinny. The layout is left-aligned and the font is not suitable for Persian text. 
[...]

PR #2855, opened 2026-03-26
feat: Custom background gradient support 

LilyLavenderPony:
### Description

This PR introduces a customizable background gradient feature, allowing users to define up to five colors, gradient angle, transparency and the option to apply the gradient on background elements only or the whole screen.

The gradient is applied on top of existing themes rather than introducing new theme variants, enabling lightweight visual customization without affecting the core theming system.

Examples using the Silver and Butter themes are shown below:

Preview2
Preview1

**Implementation**

- Introduced CustomThemeManager for dynamic background styling
- Added color pickers and controls in Appearance settings
- Gradient is computed and applied at runtime based on user settings
- Added a separate div the gradient is applied on to not interfere with existing architecture and projects of other contributors

**Design considerations**

This proposal builds on the discussion in https://github.com/cinnyapp/cinny/issues/357 with some differences:
- Blur support is implemented but currently not exposed in the UI, as its visual impact is minimal in the current layout. This can be revisited in future iterations.
- No additional structural class hooks were introduced. The implementation avoids relying on internal layout details, reducing maintenance overhead and improving compatibility with future UI changes.

**How to test**

1. - Enable "Custom Background" in settings
2. - (De-)select multiple colors
3. - Adjust angle and transparency
4. - Disable the setting and confirm it resets as intended

**Tested on:**

- Firefox
- Chrome


LilyLavenderPony:

u9000-Nine:
So adding a separate div on top isn't really the same thing as changing the background. This is especially apparent in the second image, where the gradient visibly tints all icons and text.

Also, if we're going to start adding theme options, then "Appearance" should be its own settings tab instead of dominating General.

LilyLavenderPony:
> Also, if we're going to start adding theme options, then "Appearance" should be its own settings tab instead of dominating General.

You are right, this is a good point. 

> So adding a separate div on top isn't really the same thing as changing the background. This is especially apparent in the second image, where the gradient visibly tints all icons and text.

That is true. I chose this approach because modifying the actual background would require changes that could conflict with [this PR](https://github.com/cinnyapp/cinny/pull/2769), which is currently reworking the styling. So I went with a more isolated solution for now.

However, I agree that the tinting of icons and text is unexpected behaviour for users and results in poor UX. It may be better to revisit or rework this PR once the CSS changes in Cinny have stabilized.

Thanks to you and @kfiven for the feedback, and for moving this PR to draft.

PR #2920, opened 2026-04-30
fix: edit lists crashing and list rendering issues in Firefox

nushea:
Cinny used to crash when editing a message that ends in a list with an empty item, ordered or not.

The type of list i mean:
image

With this pr it will not do so anymore.

It additionally fixes Firefox rendering lists sent from cinny with empty trailing items collapsing over each-other.

Messages sent:
abba
1. few
2. few
3. 
4. 

and respectively
few
1. few
2. few
3. 
4. 
rag

(original look on firefox/zen)
image

(new look on firefox/zen)
image

(also this pr is here because it was reported downstream and i have decided to send the fix upstream because it would be useful for anyone that happens for whatever reason to send empty lists that they would have a reason to edit later on 🤷🏻‍♀️)

### DIFF
--- a/src/app/components/editor/input.ts
+++ b/src/app/components/editor/input.ts
@@ -139,6 +139,7 @@ const getInlineElement = (node: ChildNode, processText: ProcessTextCallback): In
   }
 
   if (isTag(node)) {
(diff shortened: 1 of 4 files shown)


ajbura (approved):
Looks Good. Thanks!

PR #2923, opened 2026-05-02
Add setting to hide unread activity dot indicators

chagai95:
The idea is to get rid of showing those type of unread rooms and keep only the ones with the number which are a lot more reliable and important.

This is how it looks before:
image

and this is after:
image

### DIFF
--- a/config.json
+++ b/config.json
@@ -28,5 +28,11 @@
   "hashRouter": {
     "enabled": false,
     "basename": "/"
+  },
+
+  "translation": {
+    "endpoint": "https://mx.chagai.website/translate-api",
+    "token": "",
+    "defaultTargetLang": "en"
   }
 }
--- a/config.json
+++ b/config.json
@@ -32,7 +32,12 @@
 
   "translation": {
     "endpoint": "https://mx.chagai.website/translate-api",
-    "token": "",
+    "token": "oMBdfX9qUcwBl2TZMTNjYkC9YHuSdI2Cfne-57h8Pks",
     "defaultTargetLang": "en"
+  },
+
+  "transcription": {
+    "endpoint": "https://mx.chagai.website/apptranscribe",
+    "token": "UwTuImsqajz9PTXg7dMQb839PEPMKsaTWMRmde4V1NQ"
   }
 }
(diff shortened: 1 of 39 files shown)


chagai95:

chagai95:
recheck

chagai95:
Here's a netlify link, if you wanted to test: 
https://dev--cinny-chagai.netlify.app/home/

CrazyNicc:
Uhm, maybe I'm misunderstanding the purpose of this PR, but your "After" screenshot just seems to hide all the rooms that had this Unread Indicator

chagai95:
@CrazyNicc thx for answering so quickly!

Yes that's exactly the point. A lot of irrelevant events trigger that unread marker and I think it makes sense to hide them for some users who get a lot of messages 🙈 

Also it is shown when the room is muted I think and probably in other scenarios. There are also a lot of bugs that are related to them.

CrazyNicc:
@chagai95 Oh, afaik muting the room also removed all the read indicators.

> A lot of irrelevant events trigger that unread marker and I think it makes sense to hide them for some users who get a lot of messages

But your screenshot completely hides the rooms themselves, not just the unread markers of these rooms. Or did I miss something?

chagai95:
> @chagai95 Oh, afaik muting the room also removed all the read indicators.
> 
> > A lot of irrelevant events trigger that unread marker and I think it makes sense to hide them for some users who get a lot of messages
> 
> But your screenshot completely hides the rooms themselves, not just the unread markers of these rooms. Or did I miss something?

It hides them from the unread view, when the section is closed (arrow not down), maybe muting the room helps with that, didn't check but I don't want to start muting all my rooms just to avoid seeing this, on fluffy chat it's not shown by default in the unread tab.

chagai95:
Hi everyone, just wondering if there is something missing here?

CrazyNicc:
See the notice in the Project README:

> We are currently in the process of replacing the matrix-js-sdk with our own SDK. As a result, we will not be accepting any pull requests until further notice. Thank you for your understanding.

chagai95:
Sorry I accidently kept on using the same branch 🤦🏻

PR #2924, opened 2026-05-04
fix: support audio with ogg filetype

LeaPhant:
This adds a workaround to support playback for directly uploaded .ogg (`application/ogg`) files. Proper handling would require probing the container for all multimedia formats but WebM replaced all uses of this so Ogg video has become exceedingly rare.

<table>
<tr>
<th>Before</th>
<th>After</th>
<tr>
<td>Screenshot_26-05-04_10-00-34</td>
<td>Screenshot_26-05-04_10-01-20</td>
</tr>
</table>


ajbura (approved):
thanks!

PR #2987, opened 2026-06-12
Send attachments before the caption text

basnijholt:
## What

Inverts the room-input send-session ordering: attachments send first (in selection order) and the caption text sends last, in every send mode.

- **Existing thread / classic room mode**: same events, flipped order — uploads, then text.
- **Main timeline with text + attachments**: the `auto-thread-text-root` mode is removed. These sends now use the upload-root shape that attachment-only bursts already used: the first attachment is the room-level root, and the remaining attachments plus the caption thread under it. Reply-draft metadata stays on the root upload; the caption joins the thread with a plain thread relation.
- **Composer reset** moves from text-send time to session start (the session snapshots the caption), so the composer frees up immediately instead of staying frozen behind slow uploads. A failed caption stays in the session and retries via the existing Send-again resume path.
- A caption no longer waits on a failed non-root upload awaiting manual retry — it sends once every active upload is settled, mirroring the existing relaxed non-root retry ordering.

## Why

MindRoom's backend (mindroom-ai/mindroom#1255) coalesces a live message batch into one agent turn with a single rule: a batch ending in **text** is a complete utterance and dispatches immediately; a batch ending in **media** waits a debounce window for more attachments or the trailing caption, which closes the batch the moment it arrives.

This fork sent the events in the opposite order (CINNY-067 sent the text first so its event id could root the auto-thread). With text-first, the backend answers the caption **before the images arrive**, and the images land as a separate follow-up turn. Media-first + trailing caption makes image(s)+caption one combined turn that dispatches the instant the caption is sent.

## Test note

`RoomInput.test.ts` now also mocks `components/editor/utils`: the send-session controller imports the reset helpers from there, not from the mocked editor barrel. At the old HEAD the controller's real `resetEditor` crashed against the stubbed slate `Editor` and the send-step `catch` silently swallowed it; moving the reset to session start surfaced that latent gap.

## Validation

- `npm test`: 302/302 files, 2249/2249 tests
- `npm run typecheck` passes
- `npm run build` passes
- `npx eslint` on changed files passes
- Runbook updated (CINNY-108)


basnijholt:
Apologies — opened against upstream by mistake; this is for the mindroom-ai fork.

PR #3008, opened 2026-07-01
feat: sync ! command autocomplete with upstream MindRoom bot commands

basnijholt:
## Summary

Makes the MindRoom `!` command autocomplete in the composer mirror the actual command set of the upstream MindRoom bot (`mindroom/src/mindroom/commands/parsing.py`, `CommandType` + `_COMMAND_DOCS`), so every command the bot understands is autocompletable and no dead commands are suggested.

- **Added missing commands:** `!model [name|list|reset]`, `!thread_mode [room|thread|reset|show]`, `!reload-plugins`
- **Removed stale entries:**
  - `!skill` — explicitly removed upstream; parses as `UNKNOWN` (see upstream `test_removed_skill_command_is_unknown`)
  - `!widget` — does not exist upstream and has no client-side handler
- Entries are ordered to match the order `!help` lists them (upstream `CommandType` enum order)
- New test asserts every `syntax` string starts with `!<name>`, guarding that autocomplete-inserted text stays valid

Admin-only commands (`!reload-plugins`, `!config`, `!thread_mode`) are kept in the list since the frontend has no admin-status signal and the bot replies with a clear permission error.

## Validation

- `npm run typecheck` ✅
- `npm run build` ✅
- `npx eslint src/app/mindroom/commands/` ✅
- `npx vitest run src/app/mindroom/commands/` ✅ (6 tests)
- Independent subagent review pass against the upstream parser: verified all 10 non-UNKNOWN commands are covered, inserted text (including trailing space and hyphen in `reload-plugins`) matches the upstream regexes after `message.strip()`, and no listed command parses as `UNKNOWN`.


basnijholt:
Opened against the wrong repository by mistake — this belongs on the mindroom-cinny fork. Apologies for the noise.

PR #3020, opened 2026-07-09
fix(threads): open rooms on their room view, drop last-thread auto-restore

basnijholt:
## Why

User testing (non-technical users) showed that clicking a room and being teleported back into the last thread that was open there is disorienting — the app appears to navigate somewhere the user didn't click. This was a fork addition on top of upstream Cinny.

## What changed

Clicking a room now always lands on the room's own view (focused overview / compact / classic — whatever view mode the room is in). The whole last-open-thread persistence feature is removed end to end:

- `threads/lastOpenThread.ts` (+ test) deleted — no more per-room localStorage tracking.
- `useRoomThreadRouteRestore` keeps only its two unrelated jobs: redirecting thread URLs to the room view in classic mode, and removing a thread from recent-threads when it fails to load.
- The `ClientLayout` startup effect that re-injected the saved thread into the bare-home redirect is removed, along with its restore-only helpers in `routing/clientRouteRestore.ts` (`getLastOpenThreadRestoreTarget`, `buildThreadRestorePath`, `getRoomIdFromLastKnownPath`).
- `clientStorageAtoms` registration removed; logout cleanup now removes the legacy `lastOpenThread<userId>` localStorage key directly.

**Unchanged:** explicit thread deep links (`?threadId=...`) still open the thread; session route restore via `lastKnownPath` still works (reloading mid-thread restores the thread because the threadId is part of the stored URL); per-room view mode persistence is untouched; the swipe-based same-room thread re-entry (`lastExitedThreadAtom`) is untouched.

## Also in this PR (pre-existing breakage)

First commit fixes a case-insensitive filename collision shipped in #95: `keyBackupNudge.ts`/`.test.ts` vs `KeyBackupNudge.tsx`/`.test.ts`. On macOS/Windows the two test files collide on disk and `tsc` resolves the `KeyBackupNudge` import to the wrong module, breaking `npm run typecheck` and leaving a permanently-dirty `KeyBackupNudge.test.ts` in every checkout. The dismissal-state helpers are renamed to `keyBackupNudgeDismissal.ts` (+ test). Without this, the branch can't be validated locally on macOS.

## Validation

- `npm run typecheck` clean
- `npx vitest run` — 352 files / 2786 tests pass
- `npm run build` clean
- eslint on changed files clean
- Independent review pass (subagent): no findings

FORK_CHANGES.md Runbook updated.


basnijholt:
Apologies — this was opened against the wrong repository (intended for the mindroom-ai fork). Closing.

PR #3023, opened 2026-07-11
fix: file drop overlay staying active when drag leaves the drop zone

### LINKED ISSUE #2555: Drag & drop modal gets stuck sometimes
### Describe the bug

Sometimes the "Drop Files" gets stuck when dragging a file across the window without dropping onto it

### Reproduction

1. Drag a file across the Cinny window (don't release it onto cinny just hold it and go in and out of the window)
2. Keep going in and out of the window until the modal persists despite the file being out of the window or no longer held

Sometimes the modal gets stuck, but not always.

### Expected behavior

It should never get stuck, I tested other clients and they never get stuck

### Platform and versions

```shell
Windows 25H2
Vivaldi 7.7.3851.66 (Official Build) (64-bit)
Cinny v4.10.2
```

### Additional context

_No response_

Vladexy88x:
### Description

Fixes: #2555
The file drop overlay (`useFileDropZone`) could stay active after the dragged file left the drop zone or the window, because the previous implementation tracked drag state in a ref and only deactivated on `dragleave` when the last state was `over`.

This PR replaces the ref-based state tracking with a `relatedTarget` check in the `dragleave` handler: the overlay is deactivated when the element being entered is outside the zone, or `null` (pointer left the window). This also lets us drop the `dragStateRef` and the bookkeeping in the `drop`/`dragover` handlers.

in collaboration Claude Code

### DIFF
--- a/src/app/hooks/useFileDrop.ts
+++ b/src/app/hooks/useFileDrop.ts
@@ -1,4 +1,4 @@
-import { useCallback, DragEventHandler, RefObject, useState, useEffect, useRef } from 'react';
+import { useCallback, DragEventHandler, RefObject, useState, useEffect } from 'react';
 import { getDataTransferFiles } from '../utils/dom';
 
 export const useFileDropHandler = (onDrop: (file: File[]) => void): DragEventHandler =>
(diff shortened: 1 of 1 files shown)


Hifumorenko:
Thank you!

PR #3032, opened 2026-07-19
feat: Add a PWA badge indicator for unread messages

MarioLiebisch:
While evaluating Cinny for our team we immediately noticed there's no badge counter on the app icon, which is kind of annoying when you can't immediately check notifications all the time.


MarioLiebisch:

PR #3065, opened 2026-08-07
fix: survive third-party DOM mutation from page translation

basnijholt:
## Problem

A user on a Dutch-locale Chrome hit this on a thread at `chat.mindroom.chat`:

```
NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at pte (https://chat.mindroom.chat/assets/index-B042P1B6.js:9:26173)
    at U1 (...)
```

The whole app was replaced by react-router's default error page ("Unexpected Application Error!").

## Root cause

Chrome's page translator moves React-owned text nodes into injected `<font>` wrappers. React still holds the original node and calls `parent.removeChild(node)` on the next commit, which throws out of the commit phase and escapes to the router error boundary.

Evidence that translation was live on the document: the reported stack trace was itself translated — V8 never localizes the `at` prefix in `error.stack`, yet the screenshot shows `op pte (...)` and `bij U1 (...)`. `"Onverwachte toepassingsfout!"` also does not appear in `src/app/locales/nl.json`; it is react-router's hardcoded English heading, translated in place.

Streaming MindRoom replies send frequent `m.replace` edits, so a translated thread re-commits constantly and crashes quickly. That matches the `?threadId=` URL in the report.

App-side causes were ruled out: the only `removeChild` in `src` is `CallEmbed.ts:299`, already wrapped in a guarded `attempt()`, and `dom.ts:116` appends its temporary clipboard input outside the React root.

## Fix

**Layer one — stop the translator.** `index.html` now sets `translate="no"` on `<html>` and adds `<meta name="google" content="notranslate">`. This costs nothing in reach: the in-app language picker already covers en/de/nl, and `i18next-browser-languagedetector` already resolves `navigator` ahead of `htmlTag`, so a Dutch browser gets a Dutch UI without the browser translator.

**Layer two — survive anything else.** `installDomMutationGuard()` patches `Node.prototype.removeChild` / `insertBefore` so a wrong-parent call no-ops (or appends) instead of throwing. This covers extensions that ignore `notranslate`. It runs before the first React commit, is idempotent, returns an uninstall function for tests, and logs one `console.warn` per session so genuine app-side DOM bugs stay visible.

## Changes

- `index.html` — `translate="no"` + `notranslate` meta
- `src/app/utils/domMutationGuard.ts` — new guard
- `src/index.tsx` — install before the first commit
- `src/app/utils/domMutationGuard.test.tsx` — regression coverage
- `FORK_CHANGES.md` — Runbook entry

## Validation

- `domMutationGuard.test.tsx` (7 tests): the unguarded crash reproduces deterministically, the guarded path recovers and keeps rendering, warn-once, reparented-anchor append, well-formed calls untouched, idempotence, uninstall restoration
- `npm run typecheck` clean
- `npm run build` clean; both attributes verified present in `dist/index.html`
- `npm run lint` 0 errors, 17 warnings (existing baseline)
- `npx prettier --check` on touched files, and `git diff --check`, both clean
- `npm run test`: 3497 passed, 1 failed — `src/app/styles/scrollbarTheme.test.ts`, which reproduces on clean `dev` in a throwaway worktree. It asserts on hashed selectors in `node_modules/folds/dist/style.css` and reads no file this PR touches.

## Deploy note

Cached clients keep crashing until they pick up the new `index.html`, so this needs the Cinny deploy to `/var/www/cinny/dist` to take effect.


basnijholt:
Opened against the wrong remote — this targets our downstream fork, not upstream Cinny. Apologies for the noise.

PR #3074, opened 2026-08-19
feat: Discord-style thread side-panel (thread drawer, summaries, real unreads, media fixes)

jrimmer:
# feat: Discord-style thread side-panel

> **Note:** The repo README states PRs are closed during the SDK rewrite. This PR is opened for **visibility and reference**, not with an expectation of an immediate merge — it exists so there's a concrete, reviewable implementation of a feature that the community can see, build on, or revisit once the SDK migration settles. A public fork lives at https://github.com/jrimmer/cinny-threads for anyone who wants to run this today.

## Summary

Cinny renders threads inline in the room timeline with no dedicated panel. This PR adds a **Discord-style thread side-panel** — a drawer docked to the right of the room view (modeled on the existing `MembersDrawer`) — plus Element-style thread summaries on root messages, real unread counts in channel badges, and two build/runtime fixes that were necessary to make the feature work end-to-end.

Based on `release-v4.12.6`. 28 commits, ~2,050 lines added across 21 files.

Cinny thread side-panel

### Thread side-panel drawer (`ThreadsDrawer`)
- **Thread list** — every thread in the room, with root message preview, reply count, and a white unread-count pill. One-line rows, no sender prefix (the user is already in the channel).
- **Thread detail view** — renders the full thread conversation inside the drawer, reusing Cinny's existing message renderers (`Message`, `RenderMessageContent`, `EncryptedContent`) so it looks identical to the main timeline. Supports `m.room.message`, `m.room.encrypted` (with decrypt/retry/redacted/unsupported states), `m.sticker`, and edited messages (`m.new_content`).
- **Thread reply composer** — reply directly from the drawer with `@mention` autocomplete pills, emoji/sticker pickers, and auto-expand. Sends `m.relates_to { rel_type: 'm.thread', event_id }` correctly.
- **Resizable** — drag the left edge to resize (min 320, max 640, default 400); width persists to `localStorage`.
- **Overflow menu** — vertical-dots menu mirroring the room menu, with **Mark as Read** (threaded read receipt), **View in Thread** (focuses the root in the main timeline), and **Copy Link**.
- **Auto-mark-read** — opening a thread detail sends a threaded read receipt, clearing the unread pill and the main-timeline summary.
- **Live updates** — new threads and replies appear live without a reload; the detail view auto-scrolls to new replies.
- **Desktop only** — collapses on mobile, same `ScreenSize.Desktop` gate as the member drawer. Toggled via a threads `IconButton` in the room header, mutually exclusive with the member drawer.

### Element-style thread summaries (`ThreadSummary`)
- Thread root messages in the main timeline show a clickable summary row beneath them (thread icon + last-reply preview + reply count) when they have replies.
- When the thread has unread replies, the icon swaps to an accent thread-unread icon and a white unread-count pill appears.
- Clicking the summary opens the thread in the drawer.

### Real unread message counts (`getUnreadMessageCount`)
- The left-side channel badge now shows the **actual number of unread messages** (walking the live timeline from the read receipt to the end), not the SDK push-rule notification count — which returns 0 for "Mentions & Keywords only" rooms even when there are unread messages. Matches Element behavior.
- Respects the room notification mode: for "Mentions & Keywords only" rooms the real count is only surfaced when there is an actual mention/highlight, so the badge and desktop notifications stay quiet otherwise.

### Adaptive authenticated media (`useMediaAuthentication`)
- Cinny v4.12.6 generates authenticated v1 media URLs when the server advertises spec v1.11, but the service worker that injects the Bearer token only works in secure contexts (HTTPS / localhost). On plain HTTP LAN IPs the SW can't register, so media 401s and avatars never load.
- Gates `useMediaAuthentication` on `window.isSecureContext` so authenticated v1 URLs are only generated when the SW can actually register, falling back to unauthenticated v3 media URLs over plain HTTP. Works for both `https://chat.example.com` and `http://10.0.0.5:8083` access patterns.

### Build fix: crypto WASM glue (`vite.config.js`)
- The matrix-js-sdk crypto WASM chunk references `matrix_sdk_crypto_wasm_bg.js` at module load time, but `@rollup/plugin-wasm` only emits the `.wasm` file. Without the glue JS sibling, the chunk's top-level await rejects and the app crashes with "Failed to fetch dynamically imported module" on any route that lazy-loads the crypto SDK (e.g. post-SSO). Adds the glue file to `viteStaticCopy` targets.

## Settings

A new **Threads** toggle in *Settings → General*: "Show a threads side panel for the open room (desktop)". Off by default.

## Implementation notes

- Activates authentic SDK thread aggregation via `threadSupport: true` in `mx.startClient()` (`src/client/initMatrix.ts`); without this, `room.getThreads()` only returns threads from sync-loaded local timeline events.
- Thread list hydration uses `room.fetchRoomThreads()` (the SDK's authentic primitive — no auto-caller, the app must invoke it).
- The drawer reuses Cinny's existing message renderers with `RoomTimeline`'s context wiring so rendering stays byte-consistent with the main timeline.
- Live new-thread creation for the user's own replies subscribes to `RoomEvent.LocalEchoUpdated`, because the SDK's `handleRemoteEcho()` path skips `createThread()` (only `addLiveEvents` → `addThreadedEvents` creates threads), so without this a user's own first reply in a thread never creates the `Thread` object live and the summary doesn't appear until a reload.
- An RFC with the full design rationale lives at `docs/thread-panel-rfc.md`.

## How this was built

This feature was **vibe-coded** — implemented iteratively through an AI-assisted coding workflow using **GLM-5.2** and **DeepSeek v4 Flash**, with implementation review and structured engineering discipline provided by **[Compound Engineering](https://github.com/compound-ai/compound-engineering)** skills (plan → work → code review → test loops).

It has been **tested only in a local, private install** — not on a public deployment — against a single Synapse + MAS homeserver. It is working as designed in that environment, but has not received broad testing across homeserver configurations, client versions, or scale. Reviewers should treat it as a reference implementation rather than production-hardened.

## Test plan

- [ ] Open a room with active threads, toggle Settings → Threads on, verify the drawer appears with the thread list.
- [ ] Click a thread in the list, verify the detail view renders all replies (including encrypted and sticker events).
- [ ] Reply from the drawer composer, verify the reply appears live and auto-scrolls.
- [ ] Verify `@mention` autocomplete pills trigger in the thread composer.
- [ ] Verify the overflow menu Mark as Read / View in Thread / Copy Link all work.
- [ ] Verify resizing the drawer persists across reloads.
- [ ] Verify thread summaries appear on root messages in the main timeline and update their unread state.
- [ ] Verify channel badges show real unread counts and respect Mentions-only notification mode.
- [ ] Verify avatars/media load over both HTTPS (authenticated v1) and plain HTTP LAN (v3 fallback).

## Why a fork exists

Upstream is closed to PRs during the SDK rewrite. A public fork at https://github.com/jrimmer/cinny-threads lets people run this feature today. When upstream reopens and the SDK migration settles, the thread-panel work here can serve as a reference implementation — the component structure (a drawer modeled on `MembersDrawer`, reusing existing renderers) should translate, though the SDK calls will change.

## License

AGPL-3.0, same as upstream.

### DIFF
--- a/README.fork.md
+++ b/README.fork.md
@@ -1,0 +1,111 @@
+# Cinny Threads Fork
+
+A fork of [Cinny](https://github.com/cinnyapp/cinny) v4.12.6 that adds a
+**Discord-style thread side-panel** — the feature Cinny doesn't have.
+
+Cinny is a beautiful, Discord-style Matrix client, but it renders threads inline in
+the room timeline with no dedicated thread panel. This fork adds a proper thread
+drawer docked to the right of the room view, modeled on Cinny's existing member
+drawer, plus Element-style thread summaries on root messages in the main timeline.
+
+> **Upstream status:** Cinny is [not accepting pull requests](https://github.com/cinnyapp/cinny/issues/257)
+> while they replace the matrix-js-sdk with their own SDK. This fork is pinned to
+> v4.12.6 and maintained independently until upstream reopens. A PR has been opened
+> upstream for visibility — see the link in the PR section below.
+
+## Features
+
+### Thread side-panel drawer (`ThreadsDrawer`)
+- **Thread list** — every thread in the room, with root message preview, reply count,
+  and a white unread-count pill. One-line rows, no sender prefix.
+- **Thread detail view** — renders the full thread conversation inside the drawer,
+  reusing Cinny's existing message renderers (so it looks identical to the main
+  timeline). Supports `m.room.message`, `m.room.encrypted` (with decrypt/retry
+  states), and `m.sticker` events, plus edited messages (`m.new_content`).
+- **Thread reply composer** — reply directly from the drawer with `@mention`
+  autocomplete pills, emoji/sticker pickers, and auto-expand. Sends
+  `m.relates_to { rel_type: 'm.thread', event_id }` relations correctly.
+- **Resizable** — drag the left edge to resize (min 320px, max 640px, default 400px);
+  width persists to `localStorage`.
+- **Overflow menu** — vertical-dots menu (mirroring Cinny's room menu) with
+  **Mark as Read** (sends a threaded read receipt), **View in Thread** (focuses the
+  root message in the main timeline), and **Copy Link**.
+- **Auto-mark-read** — opening a thread detail sends a threaded read receipt, clearing
+  the unread pill and the main-timeline summary indicator.
+- **Live updates** — new threads and new replies appear live without a reload;
+  the detail view auto-scrolls to new replies.
+- **Desktop only** — collapses on mobile, same as the member drawer. Toggled via
+  a threads `IconButton` in the room header (mutually exclusive with the member drawer).
+
+### Element-style thread summaries (`ThreadSummary`)
+- Thread root messages in the main timeline show a clickable summary row beneath them
+  (thread icon + last-reply preview + reply count) when they have replies.
+- When the thread has unread replies, the icon swaps to an accent thread-unread icon
+  and a white unread-count pill appears.
+- Clicking the summary opens the thread in the drawer.
+
+### Real unread message counts in channel badges
+- The left-side channel list badge now shows the **actual number of unread messages**
+  (walking the live timeline from the read receipt to the end), not the SDK push-rule
+  notification count which returns 0 for "Mentions & Keywords only" rooms. Matches
+  Element behavior.
+
+### Adaptive authenticated media (`useMediaAuthentication`)
+- Cinny v4.12.6 generates authenticated v1 media URLs when the server advertises
+  spec v1.11, but the service worker that injects the Bearer token only works in
+  secure contexts (HTTPS / localhost). On plain HTTP LAN IPs the SW can't register,
+  so media 401s and avatars never load.
+- This fork gates `useMediaAuthentication` on `window.isSecureContext` so
+  authenticated v1 URLs are only generated when the SW can actually register, falling
+  back to unauthenticated v3 media URLs over plain HTTP. Works for both
+  `https://chat.example.com` and `http://10.0.0.5:8083` access patterns.
+
+### Build fix: crypto WASM glue
+- The matrix-js-sdk crypto WASM chunk references `matrix_sdk_crypto_wasm_bg.js` at
+  module load time, but `@rollup/plugin-wasm` only emits the `.wasm` file. This fork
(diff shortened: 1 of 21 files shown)


jrimmer:

jrimmer:
- - -
   - - -

jrimmer:
## Update: threading is now configurable (opt-out, default on)

The original threading behavior is retained for users who prefer it, with the new threading behavior being the default.

**What changed:**
- The existing `threadsDrawer` setting (Settings → General → "Show a threads side panel for the open room (desktop)") now **defaults to `true`**, so the Discord-style thread panel and the Element-style thread summaries under root messages are the default experience.
- The `ThreadSummary` component in the main timeline is now gated on that setting. Turning the setting **off** hides both the drawer **and** the summary rows, so the main timeline reverts to the exact upstream v4.12.6 appearance — thread replies still render inline as they always did, just without the added summary row.

**What's still always-on (not user-configurable):**
- `threadSupport: true` on the SDK client (`mx.startClient`). This activates the SDK's authentic thread aggregation (`fetchRoomThreads`, per-thread timeline sets). It's invisible to the user when the setting is off (no UI surfaces threads without the drawer/summaries), but it's required for the drawer to function when the setting is on. Gating it off would make the feature a no-op, so it stays on — but it has no effect on the timeline's appearance or behavior when `threadsDrawer` is false.

**Net effect:**
| `threadsDrawer` | Drawer | Main-timeline summary rows | Inline thread replies |
|---|---|---|---|
| `true` (default) | ✅ shown | ✅ shown | ✅ as always |
| `false` | ❌ hidden | ❌ hidden | ✅ as always (matches upstream v4.12.6) |

Commit: 36fa325

CrazyNicc:
The new SDK was primarily built because of a new implementation of Threads (which will come in the next major Update), so I fear you did a lot of redundant work here.

kfiven:
I wish the AI read the discussion on thread issue before doing this much work.

jrimmer:
Sadly the human read it which is why they went ahead and did the PR anyway! 😆

As I said in my initial comment "This PR is opened for visibility and reference." I didn't expect anything to actually be done with it.

I had an itch to scratch and learned a lot about Matrix and Cinny in the process. Win, win!

I'm glad to hear a bespoke threads implementation is in the offing. I look forward to playing with it.

kfiven:
It's deployed at https://fastidious-maamoul-c21df9.netlify.app (for playing only xD)

PR #3110, opened 2026-09-21
support RTL

shabane:
### Description
a `rtl text` in middle of the `ltr text` was wrong before, it shows the whole text backward.

image

Fixes #

now the message bubble show the text correcty 
image

### Checklist:

- [x] I have read and understood the [Contribution policy](https://github.com/ajbura/cinny/blob/dev/CONTRIBUTING.md).


shabane:
if you dont accectp the PR, i fully respect your choies, but you got the idea. if you can, we are happy to have the RTL thing fixed.

thank you so much.

